PR #239, opened 2014-01-12
Keep js on calendar to prevent conflicts with other plugins/themes.

cojennin:
The following should solve this problem http://wordpress.org/support/topic/mail-poet-former-wysija-newsletter-conflict/ and other issues that might arise from Edit Flow js getting enqueued on the wrong pages.


cojennin:
Closing in favor of #351

Issue #243, opened 2014-01-27
iCalendar feed doesn't use gmt date for start/end


mschierberl:
Publish times are reported incorrectly when subscribing to the iCalendar feed in Google Calendar or Apple Calendar.

Times appear to be adjusted by both the local time on the server and local time in the user's calendar.  iCalendar start/end times should use UTC times.

PR #244, opened 2014-01-27
Use GMT time for start and end dates in iCalendar feeds.

mschierberl:
Fixes Issue #243

Use post_date_gmt for start and end date in iCalendar feeds.


justnorris:
Closing in favor of #405

PR #259, opened 2014-05-14
Only subscribe to actions in "admin mode"

jonathan-rodan:
The plugin doesn't need to load all of the modules if the user is not in an admin page.

Changing this improves the user experience while looking at articles.

The profiler was assigning 8% of the read article time to this plugin. (The profiler was run only on views to the content)

**_BEFORE:**_
screen shot 2014-05-14 at 6 34 28 pm

**_AFTER**_:
screen shot 2014-05-14 at 6 26 33 pm

### DIFF
--- a/edit_flow.php
+++ b/edit_flow.php
@@ -142,12 +142,14 @@ class edit_flow {
 	 * @uses add_action() To add various actions
 	 */
 	private function setup_actions() {
-		add_action( 'init', array( $this, 'action_init' ) );
-		add_action( 'init', array( $this, 'action_init_after' ), 1000 );
+        if(is_admin()) {
+            add_action( 'init', array( $this, 'action_init' ) );
+            add_action( 'init', array( $this, 'action_init_after' ), 1000 );
 
-		add_action( 'admin_init', array( $this, 'action_admin_init' ) );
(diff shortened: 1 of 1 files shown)


cojennin:
Seems pretty risky to change up the loading strategy without any tests to ensure this doesn't cause unexpected issues or regressions. Especially since the [setup_actions](https://github.com/Automattic/Edit-Flow/blob/master/edit_flow.php#L145) function already seems to indicate when certain functionality should be loaded on [admin_init](https://github.com/Automattic/Edit-Flow/blob/master/edit_flow.php#L148) vs [init](https://github.com/Automattic/Edit-Flow/blob/master/edit_flow.php#L145).

If someone can definitively ensure that Edit Flow should only be using `admin_init`  instead of `init` inside of [setup_actions](https://github.com/Automattic/Edit-Flow/blob/master/edit_flow.php#L145) I'd be willing to merge this patch, but if not I think that investigation needs to be done first. Feel free to take it on, I imagine it would include both tests and an examination of what [setup_actions](https://github.com/Automattic/Edit-Flow/blob/master/edit_flow.php#L145) calls to ensure `admin_init` can be used without issue.

Issue #270, opened 2014-07-23
Indicate who was notified when an editorial comment (or other notification) is sent


jerclarke:
A big issue our users are having is that after they submit an editorial comment they don't remember if they subscribed a user/group in Notifications before or after the comment. 

It would be nice if there was some indication after submitting a comment as to who was notified. If they saw a message like "No one was notified" it would remind them to add their editor in Notifications and add a new comment/re-edit the status. If they saw "Sub-Editors Group notified" then they could rest easy. 

Ultimately the same goes for status changes (a message at the top below "post saved") and everything else. Users would get peace of mind knowing the notifications were sent and eventually would learn to miss the messages when they had failed to subscribe the appropriate person. 

Any leads on the most efficient way to build this as a plugin are very welcome.

sboisvert:
@WPprodigy I think this could be a good issue to work on. Do you want to test it and see if it works as intended and work with @jerclarke and @goodguyry on if it fits the plugins workflow?

WPprodigy:
Hey @goodguyry,

Thanks for the PRs, nice work! Just checking in to see if you think you’ll have some time to work on this in the near future? Happy to help assist with this further. If you don’t think you’ll be able to get around to this soon, just let me know and I can take over the PRs from here if you would like.

jerclarke:
FWIW: This ticket is pretty much done, as the main change is already merged to `master`. 

If someone can merge my change to the email template in #479 (as described in #478) then this feature will be really really finished (since I think they are related). Both could probably go into `0.8.3` together if there's still will to do that release before moving to `0.9`.

jerclarke:
Closing this because it's already been merged, and remaining fixes/improvements (e.g. #483 ) are something else.

PR #271, opened 2014-07-30
Adds follower feedback in comments/replies as part of #270

goodguyry:
Indicates whether or not any users or user groups were selected to receive a notification of a new comment (the first part of #270). Feedback is given both in the comment/reply form and in the comment itself (on hover, like the reply link). The indicator in the comment persists, for transparency's sake, when/if Edit Flow Notifications are turned off.

I split #270 up into two pull requests - I believe them to be two separate features.

### DIFF
--- a/modules/editorial-comments/editorial-comments.php
+++ b/modules/editorial-comments/editorial-comments.php
@@ -200,7 +200,7 @@ class EF_Editorial_Comments extends EF_Module
 			<?php
 			if ($this->module_enabled( 'notifications' )) {
 			// Only show the input if notifications are enabled ?>
-			<label for="ef-reply-notifier">The following will be notified:
+			<label for="ef-reply-notifier"><?php _e('The following will be notified:', 'edit-flow'); ?>
 				<input id="ef-reply-notifier" class="ef-reply-notifier-message" readonly>
 			</label>
 			<?php } ?>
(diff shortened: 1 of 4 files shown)


cojennin:
@goodguyry would you still be interested in working on this PR? Seems like useful functionality to have. Would be great to refine this PR a bit and get the tests working again.

goodguyry:
Sure. I'll find time to get reacquainted and see what I can do.

cojennin:
@goodguyry Awesome, let me know if you need a hand with getting reacquainted!

WPprodigy:
Did a quick review and some testing. All seems to be working good still and as intended 👍 . I'm going to close out this PR shortly and open up a new one to build on top of what's been done here - just posting a couple comments below for what stuck out that will need to be done.

### editorial-comments.php

`$_POST['notification']` needs to be sanitized before saving to the db in comment meta. Also should check if it `isset` first before accessing.

Will need to escape the output in the `get_comment_notification_meta` function since `$notification` is technically coming from an external place (the db).

To better fit into the current templating, it would be best to rename the above function to `comment_notification_message` and just return the string. Then the HTML markup with all the other HTML where the function is called.

Would probably be good to introduce a filter to disable this new functionality as well, in case it won't be useful to a site and they don't want extra unneeded DB entries.

Lastly, instead of `1` being the default/empty value for the comment meta - I'm thinking it would be best to just not add comment meta at all. Then we can just check if the response is empty/false when trying to retrieve it. This will help prevent unnecessary DB additions.

_____

### editorial-comments.js

Checking if the value is `No one will be notified` won't work out if the string is translated, so would be good to avoid needing to do that. Also ideally the default value won't be `1` but rather just empty. So this conditional will be able to be simplified.

The `and` in the list of subscribers along with the `No one will be notified` text will need to be translatable. So will need to pass them through on the PHP side when the script is localized.

Would be good to avoid `.html()`, so instead we can just be a bit more specific about where the text is being pulled from.

Minor thing, but the property `notify` seemed a bit confusing to me at first, not sure if that verb is a good fit for what the function is doing. Considering renaming to perhaps `notifiedSubscribers` or `notifiedMessage`. Maybe just to picky on my part though though ;)

### notifications.js

Rather than initiating the function in this file, I'm thinking it would be better to keep it all together in `editorial-comments.js` and instead just watch for notification settings changes. To do this, could perhaps just emit a custom event and watch for that to be called.

WPprodigy:
Closing in favor of https://github.com/Automattic/Edit-Flow/pull/452

PR #272, opened 2014-07-30
Adds follower feedback in post submit box as part of #270

goodguyry:
Indicates whether or not any users or user groups are following a page/post (the second part of #270). Feedback is given in the post submit box, based on the users and groups selected in the Edit Flow Notifications meta box.


WPprodigy:
Closing here and continuing in https://github.com/Automattic/Edit-Flow/pull/453. Thanks for the contribution!

PR #278, opened 2014-10-02
Change to event delegation for ef click event

sanchothefat:
I had a problem in WP 3.9.2 while using chrome where the checkbox selector was running incredibly slowly and incorrectly adding event handlers to the extent that the post list admin pages were unusable for up to 30 seconds.

html-parse-timings

Changing line 9 to use `.on()` fixed it.

Possibly to do with changes in the sizzle library but this fixes it.

### DIFF
--- a/modules/custom-status/custom-status.php
+++ b/modules/custom-status/custom-status.php
@@ -415,12 +415,14 @@ class EF_Custom_Status extends EF_Module {
 
  			$post_type_obj = get_post_type_object( $this->get_current_post_type() );
 			
+			$default_custom_status = $this->get_default_custom_status()->slug;
+			
 			// Now, let's print the JS vars
 			?>
 			<script type="text/javascript">
 				var custom_statuses = <?php echo json_encode( $all_statuses ); ?>;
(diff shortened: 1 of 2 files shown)


sanchothefat:
Yeah agree on that count - good to bind as low as possible, maybe on the
list table itself or just above that?

-- 
Robert O'Rourke
*mobile:* +447816329424
*skype: *rob.o.rourke
*web:* röb.co

cojennin:
Looks like the JS changes have already [been merged in](https://github.com/Automattic/Edit-Flow/blob/master/modules/notifications/lib/notifications.js#L9).

There's a couple of other changes in this PR [regarding custom statuses](https://github.com/Automattic/Edit-Flow/pull/278/files#diff-228291de10e40133681b1ff8414a9a54R418). Do you mind splitting those out into a separate PR with a few comments on the modifications being made?

sanchothefat:
@cojennin sure, bad habit of mine :/ Looking at those changes I don't think they're really necessary now. Will close this PR

cojennin:
@sanchothefat heh, no worries. Thanks for checking back in on this PR!

PR #287, opened 2014-10-30
Fix performance issue for Chrome http://stackoverflow.com/questions/4693...

idimon:
We face to conflict with Bulk Press plugin. The main reason was low performance for Google Chrome browser (we have 2000+ checkboxes on page) http://stackoverflow.com/questions/4693809/chrome-hangs-when-using-javascript-to-check-many-checkboxes-while-firefox-does.


dcpedit:
I am having this issue as well (page has 1600 checkboxes).  I've tested the fix from @idimon and the performance is greatly improved in Chrome after patching

PR #292, opened 2014-11-21
List of notification users generated via AJAX.

mariusveltan:
Moved the generation of the list of users to be notified into an asynchronous call. By generating the list at page load, synchronously, the browser window ends of freezing for a couple of seconds, in case the list of users has a high number of  entries.


cojennin:
Do you have some more context on when this issue is triggered? Does the page freeze when there are too many users and user groups?

Making this an ajax call seems like a bit of overkill, but I'd like to hear more about the problem before we think up some alternatives.

cojennin:
Just following up on this PR. Would you be able to provide some more details on the issue? How many users do you have for your site?

mariusveltan:
Unfortunately, I can no longer offer specific details regarding the context in which this PR was necessary.

cojennin:
@mariusveltan no worries! 

Going to close this out for the time being. I'll make sure to reference this PR if I hear about this issue from other folks. Thanks!

Issue #295, opened 2014-12-09
Calendar module's JS interferes with another plugin that uses an .editing class


daveross:
reset_editorial_metadata() is hiding elements from another plugin we're using. The selector $('.editing') is overly-broad. I'll submit a PR to narrow the focus to just look at the Calendar page, but you may want to address this closer to where the click handler is defined instead.

philipjohn:
Closing in favour of 351 which covers the issue in a little more detail.

PR #296, opened 2014-12-09
Tighten the scope of search for .editing elements in the page

daveross:
reset_editorial_metadata() is hiding elements from another plugin we're using. The selector $('.editing') is overly-broad. This PR narrows the focus of the selector to just look on the Calendar page, although you might want to address this conflict differently (closer to the click handler?).

Addresses issue #295

### DIFF
--- a/modules/calendar/lib/calendar.js
+++ b/modules/calendar/lib/calendar.js
@@ -43,7 +43,7 @@ jQuery(document).ready(function ($) {
 
 
 	function reset_editorial_metadata() {
-		$('.editing').removeClass('editing')
+		$('.dashboard_page_calendar').find('.editing').removeClass('editing')
 			.addClass('hidden')
 			.prev()
 			.removeClass('hidden');


cojennin:
Agreed, these should be scoped more appropriately to prevent clashing with other plugins. Seems like you'll still run into issues in [other places](https://github.com/Automattic/Edit-Flow/blob/master/modules/calendar/lib/calendar.js#L85) where [broader selectors](https://github.com/Automattic/Edit-Flow/blob/master/modules/calendar/lib/calendar.js#L88) are used. 

Do you have some time to work on scoping a [few of these](https://github.com/Automattic/Edit-Flow/blob/master/modules/calendar/lib/calendar.js#L69) selectors within `calendar.js` to be more specific?

Also agree with addressing these issues closer to the click handler (might be possible to leverage event delegation on a few?).

cojennin:
Just wanted to follow up on this PR. If you have some time to take a crack at it that would be great!

cojennin:
Think you'll be able to finish out this PR with some of the suggested tweaks above? If not I'm going to recommend closing. Definitely would be great to get this work in but I'd like to make this a consistent and conscious change.

Let me know!

cojennin:
Going to close this out. Think the Edit Flow javascript could definitely use some cleanup work around scoping, so I'll leave #295 open. Would be great to see a patch that tackles all (or most of) the scoping issues in the calendar javascript

PR #298, opened 2015-02-03
Add default value functionality, fixes #276

### LINKED ISSUE #276: Check box custom field - default to checked
On load of a new custom post, I like to be able to default the check box custom field to "checked".  Currently I use java script but find it not as reliable and from time to time causes issues and conflict with other plugin.  It would be great if we can build an option to default this.

chances:
In addition to fixing #276, this PR also presents default value functionality for all current editorial metadata field types.

### DIFF
--- a/modules/editorial-metadata/editorial-metadata.php
+++ b/modules/editorial-metadata/editorial-metadata.php
@@ -883,6 +883,7 @@ class EF_Editorial_Metadata extends EF_Module {
 				'slug' => $old_term->slug,
 				'description' => $old_term->description,
 				'type' => $old_term->type,
+                'default' => $old_term->default,
 				'viewable' => $old_term->viewable,
 			);
 		$new_args = array_merge( $old_args, $args );
@@ -892,6 +893,7 @@ class EF_Editorial_Metadata extends EF_Module {
 			'description' => $new_args['description'],
 			'position' => $new_args['position'],
 			'type' => $new_args['type'],
+            'default' => $new_args['default'],
 			'viewable' => $new_args['viewable'],
 		);	
 		$encoded_description = $this->get_encoded_description( $args_to_encode );
@@ -923,6 +925,7 @@ class EF_Editorial_Metadata extends EF_Module {
 			'slug' => '',
 			'description' => '',
 			'type' => '',
+            'default' => '',
 			'viewable' => false,
 		);
 		$args = array_merge( $defaults, $args );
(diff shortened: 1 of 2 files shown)


cojennin:
Editorial Metadata is a nice overlay on top of regular post meta so you should be able to run add_post_meta with `unique` set to true when you create a new post to set the default value. Letting user's set defaults isn't a bad idea, but seems like it'll add extra work we'll need to think through first.

Thoughts?

chances:
Ah, and I had gone ahead and reinvented the wheel setting up the default value. Whoops.

Perhaps Edit-Flow can setup the `add_post_meta` itself for default values? I can go ahead and refactor the code to use `add_post_meta` if you'd like. But, if you still need to do some designing I can hold off.

Regarding the usefulness of default values, I had implemented this specifically for use by my previous employer. They specifically needed a few fields to have default values. Judging by #276, I'm not the only one who needed this kind of behavior. I figured it best to contribute my changes back upstream rather than relying on the monkey patch.

cojennin:
Definitely appreciate you taking the time to do the contribution! 

Couple things I'm wondering:
- Do you remember which fields were supposed to be set by default? 
- Was the desire for setting these fields to some default due to author's not filling them in?

chances:
I remember they needed a checkbox to be checked by default.

Specifically, it was a "Needs photo" verification for their photographer to go through and see which posts still needed a photograph. And yes, the authors were supposed to check the box if a new post needed a photo, but they usually forgot. With the default value, it was simpler for the photographer to go through post drafts later and verify that posts had photos.

cojennin:
@chances Gotcha. Seems like it's been a while, but curious to know: did folks toss around the idea of creating a checkbox editorial metadata called "Has Photo"? I'd imagine a photographer could have gone through all the posts where it was unchecked and check it as they went along.

Would be great to get your thoughts on the upsides and downsides of that approach (and whether you folks considered that as an alternative to implementing the "default checked")!

cojennin:
@chances let me know if you have any further thoughts (see last comment)! Otherwise I'll go ahead and close this PR. I could see some use cases for default functionality, but I I'd like to dive deeper into the workflows of folks requesting it before implementing it.

Appreciate the contribution and your [comment above](https://github.com/Automattic/Edit-Flow/pull/298#issuecomment-226957004) is a great to start for investigating the use case for this functionality!

chances:
@cojennin No, to my knowledge, we did not discuss the possibility of using the inverse "Has photo" metadata. In hindsight that seems like the better alternative in the interest of minimizing complexity and not reinventing the wheel.

cojennin:
@chances interesting, thanks for the follow up! Going to close this out, glad to get the chance to think through this a bit!

PR #300, opened 2015-02-14
gitignore: remove edit-flow.php

Ramoonus:


joshbetz:
Can you add some more details explaining this change?

See: http://vip.wordpress.com/documentation/commit-messages/

Ramoonus:
the main file is on Git ignore, which means it doesn`t track versions

joshbetz:
The main file is edit_flow.php (vs edit-flow.php)

Issue #301, opened 2015-02-20
Publish button shows instead of Schedule button for posts scheduled in the future


samglover:
Here is an example:

![image](https://cloud.githubusercontent.com/assets/2208482/6307007/ddf4eb72-b8fd-11e4-8f3f-8b15b196958b.png)

If you click the button, it does what it says and publishes the post immediately instead of scheduling it.

kellykarnetsky:
We're having a similar issue but the button still says Schedule but publishes the post immediately. I have found others having the same issue as well here https://wordpress.org/support/topic/draft-posts-being-published

samglover:
That's actually happening to us as well. I think it's probably the same bug.

kellykarnetsky:
I'm thinking the same thing as well. I'm going to look into as soon as I get a chance this week or next but hopefully someone from Automattic will step up to look since it can be used by WP VIP sites.

natebot:
I have heard reports from my team about the same behavior. I'd like to take
a look at this but likely won't have time until late next week.

kel:
I'm looking into this right now but I haven't found a good way of reproducing it. I've spent the last couple of hours going through our workflow and testing different combinations and haven't been able to reproduce it again. 

### Situation 1
* Editor A creates a post and saves it as a draft.
* Editor A schedules post for future date.
* Editor B makes changes and observes future scheduled date.
* Editor B clicks the schedule button.
* Editor B notices that post is published immediately. 

### Situation 2
* Editor A creates a post and saves it as a draft.
* Editor A schedules post for future date.
* Editor B makes changes and observes future scheduled date.
* Editor B observes publish button.
* Editor B has to change the scheduled datetime in order for schedule button to reappear.

samglover:
Try creating the post from the calendar.

On Monday, April 20, 2015, Kelly Karnetsky <notifications@github.com> wrote:

> I'm looking into this right now but I haven't found a good way of
> reproducing it. I've spent the last couple of hours going through our
> workflow and testing different combinations and haven't been able to
> reproduce it again.
> Situation 1
>
>    - Editor A creates a post and saves it as a draft.
>    - Editor A schedules post for future date.
>    - Editor B makes changes and observes future scheduled date.
>    - Editor B clicks the schedule button.
>    - Editor B notices that post is published immediately.
>
> Situation 2
>
>    - Editor A creates a post and saves it as a draft.
>    - Editor A schedules post for future date.
[...]

PR #305, opened 2015-04-28
Updating custom status hack to prevent publishing when scheduling

kel:
This is a fix for ticket #301. After setting a scheduled datetime for a post and saving it as a custom status and the user makes any changes the autosave fires. This calls the `fix_custom_status_timestamp` function that has a conditional to blank out `$data['post_date_gmt']` if `$_POST['aa']` is empty. `$_POST['aa']` is not passed on autosaves causing WordPress to publish the current post when clicking schedule.

### DIFF
--- a/modules/custom-status/custom-status.php
+++ b/modules/custom-status/custom-status.php
@@ -1327,7 +1327,7 @@ class EF_Custom_Status extends EF_Module {
 		if ( $this->disable_custom_statuses_for_post_type()
 			|| ! isset( $edit_flow )
 			|| empty( $_POST ) 
-			|| ( defined( 'DOING_AUTOSAVE' ) && DOING_AUTOSAVE ) ) )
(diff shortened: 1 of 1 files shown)


kel:
Looks like there is an issue with HHVM causing those tests to fail.

https://github.com/facebook/hhvm/issues/5215
https://github.com/travis-ci/travis-ci/issues/3790

FewKinG:
Confirmed this bug, took me some time to find out what was happening :)
The suggested solution in this commit didn't work for me because I am calling wp_insert_post inside a wordpress ajax call which actually contains $_POST data.
One thing that did work was unsetting the $_POST variable after I was done with it and before calling wp_insert_post:
`unset ( $GLOBALS[ '_POST' ] );`
But I don't think that should be the way to go...

cojennin:
Closing, should be fixed by #343.

PR #307, opened 2015-05-08
Javascript performance

hyperionjrw:
Fix for current version of jQuery and the "All posts" page. Would
previously hang on page load with current versions of WordPress.


Tarendai:
Thanks for the optimisation!

Issue #309, opened 2015-05-25
Add Composer support.


ryanmarkel:
Would like to add Composer support for the plugin - I went to create a PR but noticed that master is currently tagged in the plugin header as 0.8.1-alpha; was unsure if those changes are considered stable.

Tarendai:
So long as we tag versions with the right names they'll be interpreted by composer correctly, 0.8.1-alpha shouldn't be seen as stable by composer

spacedmonkey:
I created a pull request with a composer file. Should be simple enough to merge, it is a none breaking change. Don't see what the big deal is....

ryanmarkel:
Solved in #310.

spacedmonkey:
Thanks @ryanmarkel . As you are on the automattic organization, can you help a brother of with some similar PR I have for other automattic's repos. 

https://github.com/Automattic/advanced-post-cache/pull/2
https://github.com/Automattic/nginx-http-concat/pull/8
https://github.com/Automattic/Rewrite-Rules-Inspector/pull/16

Thanks again. :smile:

PR #310, opened 2015-06-12
Composer support

spacedmonkey:
By adding this file, it adds support for the popular php dependancy manager composer.


joshbetz:
See also #309

PR #325, opened 2016-01-23
Proposed fix on Calendar module for "current_user_can" capability check

keg:
If a custom user role has been created for write access to the calendar via the filter on https://github.com/Automattic/Edit-Flow/blob/master/modules/calendar/calendar.php#L87 and that user doesn't have WP `publish_posts` capabilities, they won't be able to create calendar entries. Proposing changing the check from `publish_posts` to `$this->create_post_cap`


cojennin:
Makes sense to me. We're using `$this->create_post_cap` to [verify the ajax call](https://github.com/Automattic/Edit-Flow/blob/master/modules/calendar/calendar.php#L1505). Should be using it on the front end as well.

Will verify it works and merge.

cojennin:
Looks good, thanks!

keg:
👍

PR #328, opened 2016-03-09
fix initialization of $post_int in EF_Module->get_current_post_type()

jeremyclarke:
It was a line that just said "$post_int;" which does nothing AFAIK.

In Debug Bar I was seeing this error:
 NOTICE: /class-module.php:243 - Undefined variable: post_int

The whole method could use a review but this PR is the minimum possible change to fix the notice and (I think) honor the original intention of the code. 

Thanks!

### DIFF
--- a/common/php/class-module.php
+++ b/common/php/class-module.php
@@ -253,7 +253,7 @@ class EF_Module {
 	function get_current_post_type() {
 		global $post, $typenow, $pagenow, $current_screen;
 		//get_post() needs a variable
-		$post_int;
+		$post_int = '';
 		if( isset( $_REQUEST['post'] ) )
 			$post_int = (int)$_REQUEST['post'];
 


common/php/class-module.php
```diff
@@ -253,7 +253,7 @@ function enqueue_datepicker_resources() {
 	function get_current_post_type() {
 		global $post, $typenow, $pagenow, $current_screen;
 		//get_post() needs a variable
-		$post_int;
+		$post_int = '';
```
mjangda: Can probably simplify to this:

```
$post_int = isset( $_REQUEST['post'] ) ? (int)$_REQUEST['post'] : false;
```

In any case, we need to change one of the `elseif` blocks below:

```
&& isset( $_REQUEST['post'] )
 			&& isset( $post_int )
```

to

```
&& $post_int
```

jeremyclarke: Sure! Figured I'd let you fix it up if you wanted. As long as the notice goes away I'm happy :)

cojennin:
@jeremyclarke do you have some time to make the finishing touches on this PR per @mjangda's comments? Should be good to merge after they're made.

cojennin:
Closing per #345

PR #331, opened 2016-04-05
Adding $post param to preview_post_link filter to prevent warning.

micahwave:
The `$post` param was added to `preview_post_link` in 4.0. Without it here it will throw warnings for others that are using the filter and expecting it to be present.


cojennin:
Looks like that filter is also applied [in the calendar](https://github.com/Automattic/Edit-Flow/blob/master/modules/calendar/calendar.php#L931). If you can fix it there as well (and anywhere else I might have missed) I think we'll be all set to merge this PR.

cojennin:
@micahwave looks good, thanks!

PR #330, opened 2016-03-24
Corrections in brazilian portuguese translation

angelosds:

PR #335, opened 2016-04-14
Remove Japanese Translation

kraftbj:
The WordPress.org translation is at 100%, so this language will be delivered via language pack. No need to bulk up the plugin with the MO/PO.


cojennin:
Should outstanding and future translations (ex: #330) be directed to https://translate.wordpress.org?

robertsky:
@cojennin i second that.

kraftbj:
I defer to someone on VIP (I'm on a different team), but Jetpack is pushing to translate.w.org, so I third. :)

cojennin:
I like the idea of utilizing these language packs and it seems like it's going to be [the expected way of delivering translations for plugins on .org](https://make.wordpress.org/meta/2015/07/10/plugins-themes-and-translate-wordpress-org/).

I would suggest the following steps:
- Until a language is 100% on [https://translate.wordpress.org](https://translate.wordpress.org/) it won't be removed from this Github project
- Until a language is 100%, we'll continue to accept PR's regarding translation improvements but ask that the PR author also provide the translation improvements on [https://translate.wordpress.org/](https://translate.wordpress.org/) prior to merging (my preference would be to not enforce this rule with an iron fist since I'd rather see the translation merged without the corresponding translation on .org then to not be merged at all. More guideline than rule)
- Figure out a strategy and expected timeline for pushing to SVN trunk to handle [updating translations for new releases](https://make.wordpress.org/plugins/2015/09/01/plugin-translations-on-wordpress-org/#comment-42293). I'm still unsure how this is handled (if a new version of EditFlow is released but the translations for it aren't at 100%, does it fall back on using the last version of the translations?). Seems like without a critical mass of folks willing to do translations this could be problematic

robertsky:
@cojennin 

> Figure out a strategy and expected timeline for pushing to SVN trunk to handle updating translations for new releases. I'm still unsure how this is handled (if a new version of EditFlow is released but the translations for it aren't at 100%, does it fall back on using the last version of the translations?). Seems like without a critical mass of folks willing to do translations this could be problematic

In the same post, the process is described in [another comment](https://make.wordpress.org/plugins/2015/09/01/plugin-translations-on-wordpress-org/#comment-42283). tl;dr: if the current version is 1.3 with 80% translated, the previous version with 100% translated will be used.

cojennin:
@robertsky good spot, thanks for the clarification. Seems like we're good to remove this translation then. Only thing I think we'll need to work out is when to sync with svn trunk for handling updates to translations for new versions.

robertsky:
@cojennin, how about sync to the development branch say a week before the milestone for the next version closes? This would give translators some time work their magic before going live.

Also, I suggest starting from somewhere for the rest of translation. i.e. shifting the translations from here to glotpress. I am currently translating Simplified Chinese. However if time permits, I will do just that for the other languages.

PR #336, opened 2016-04-20
Updating travis to test for php7.0 support.

cojennin:
Exclude hhvm and php7.0 on oldest supported versions of WordPress. Tweaking CI to test for currently stable supported version of WordPress, last stable version of WordPress and earliest supported version of WordPress with PHPUnit (3.7).

Basing this cleanup work on the [backupwordpress](https://github.com/humanmade/backupwordpress/blob/master/.travis.yml) Travis CI yaml.


.travis.yml
```diff
@@ -6,16 +6,31 @@ php:
...
     - WP_VERSION=latest WP_MULTISITE=0
     - WP_VERSION=latest WP_MULTISITE=1
-    - WP_VERSION=4.1 WP_MULTISITE=0
-    - WP_VERSION=4.1 WP_MULTISITE=1
-    - WP_VERSION=4.0 WP_MULTISITE=0
-    - WP_VERSION=4.0 WP_MULTISITE=1
-    - WP_VERSION=3.9 WP_MULTISITE=0
-    - WP_VERSION=3.9 WP_MULTISITE=1
+    # Current stable
+    - WP_VERSION=4.5 WP_MULTISITE=0
```
joshbetz: I think we should leave this on `latest` so it updates if we forget.

.travis.yml
```diff
@@ -6,16 +6,31 @@ php:
...
+    - WP_VERSION=4.5 WP_MULTISITE=1
+    # Previous Stable
+    - WP_VERSION=4.4.2 WP_MULTISITE=0
+    - WP_VERSION=4.4.2 WP_MULTISITE=1
+    # earliest supported that includes phpunit tests in core WP
+    - WP_VERSION=3.7 WP_MULTISITE=1
```
joshbetz: I'd only test the last 2 stable versions.

joshbetz:
👍 to testing PHP7

.travis.yml
cojennin: Should still be testing latest per https://github.com/Automattic/Edit-Flow/pull/336/files#diff-354f30a63fb0907d4ad57269548329e3R12, right?

.travis.yml
cojennin: Agreed. Last stable and also latest version is 4.5. Could bump that to 4.4.2 and then bump 4.4.2 down to 4.4.1 or 4.3.3?

.travis.yml
joshbetz: @cojennin Oh, you're right. Should we test latest and 4.5 explicitly? I don't feel super strongly about it, but tests will run faster without duplication.

cojennin: Agree with you on avoiding duplication. Bumping the current stable to 4.4.2 and previous stable to 4.4.1. Keeping latest as is.

kraftbj:
Any thought to how long to keep backwards compat with older versions of WP? Jetpack maintains current-1 (we support back to 4.4 since 4.5 is out, will bump to minimum 4.5 when 4.6 ships, etc). It would cut down on the poor load we ask our friend Travis to handle and potentially minimizes compat code we'll have to add as things are added.

cojennin:
@kraftbj definitely something to think about. Especially since it doesn't look like PHPUnit tests were introduced to develop.svn.wordpress.org until [3.7](https://develop.svn.wordpress.org/tags/3.7/tests/phpunit/). [Claiming support for 3.5](https://wordpress.org/plugins/edit-flow/) without testing for it seems a little disingenuous.

What about trimming the php versions we have in the yaml? Looks like Jetpack [only uses 5.6?](https://github.com/Automattic/jetpack/blob/master/.travis.yml#L8). I'd be comfortable with using just 5.6, 7.0 and hhvm and removing 5.3, 5.4 and 5.5.

mjangda:
I'd suggest testing all supported versions of PHP (i.e. `5.2 => 7` + hhvm) but dropping WP support to `current-1` like @kraftbj noted.

kraftbj:
Jetpack tests a few different variants: https://github.com/Automattic/jetpack/blob/master/.travis.yml#L26

We'd try to find a balance between not spending a ton of time testing every possible variation (5.2, 5.3, 5.4, 5.5, 5.6, 7, WP-1, WP, WP trunk) while still having fair coverage. We test on hhvm and PHP nightly (currently 7.1), but allow failures on both even though it they almost always pass.

cojennin:
@kraftbj made a few changes to emulate Jetpack's travis config, do you mind taking a look when you get a chance? Think that should give fairly comprehensive coverage while keeping the travis workload lean.

Replaced the script that was being used with a [few lines in the](https://github.com/Automattic/Edit-Flow/pull/336/files#diff-354f30a63fb0907d4ad57269548329e3R54) in the `before_script` section of the yaml as well. Makes it easier to understand the testing setup.

cojennin:
Unsure why we're [getting this failure](https://travis-ci.org/Automattic/Edit-Flow/jobs/124768311) now. I can take a look into whether modifying the testing setup started triggering it.

cojennin:
Don't have a whole lot of familiarity with Travis or PHPUnit. We're we previously not running standard WP Multisite tests? Not seeing the same output regarding multisite testing between [an older Travis build](https://travis-ci.org/Automattic/Edit-Flow/jobs/123148669) and a build with [changes to the travis testing setup](https://travis-ci.org/Automattic/Edit-Flow/jobs/124768311).

cojennin:
Alright, sorted it all out. PR should be good for review if someone has a few minutes to check it out.

cojennin:
If no objections, going to merge this shortly

kraftbj:
Apologies for the delay from me. AFK and traveling and whatnot. It looks good to me and Travis likes it.

cojennin:
No worries.

Merging.

Issue #318, opened 2015-07-29
Swedish translation


Warpsmith:
I don't know how Github works and don't know how to add files. But I've made a Swedish translation. Maybe someone who know Github can add it? Or not. It can be found here:

https://dl.dropboxusercontent.com/u/15071513/edit-flow-sv_SE.zip

cojennin:
@Warpsmith Thanks for the translation! It's been merged in per #338. We're looking to use the .org provided language packs for future versions of Edit Flow. If you have some time and can [help with the translation on translate.wordpress.org](https://translate.wordpress.org/locale/sv/default/wp-plugins/edit-flow) that'd be great!

cojennin:
Closing this issue. @Warpsmith let me know if you have some time to work on the translation on [translate.wordpress.org](https://translate.wordpress.org/locale/sv/default/wp-plugins/edit-flow). Thanks again!

Warpsmith:
@cojennin I'm on tour at the moment, but I think I have time in June-July to do some translations.

PR #338, opened 2016-04-22
include Swedish translation as provided by @Warpsmith in #318.

robertsky:


cojennin:
Awesome, thanks for implementing this as a PR @robertsky and thanks @Warpsmith for the translation!

Issue #321, opened 2015-09-18
Only variables should be passed by reference


mihdan:
`wp-admin/post.php?post=162634&action=edit`

```
Only variables should be passed by reference - /wp-content/plugins/edit-flow/modules/custom-status/custom-status.php:638 
#6 /wp-content/plugins/edit-flow/modules/custom-status/custom-status.php:422 - EF_Custom_Status->get_default_custom_status()
#5 [internal call] - EF_Custom_Status->post_admin_header('')
#4 /wp-includes/plugin.php:503 - call_user_func_array(Array[2], Array[1])
#3 /wp-admin/admin-header.php:120 - do_action('admin_print_sc...')
#2 /wp-admin/edit-form-advanced.php:413 - require_once('/var/www/roomb...')
#1 /wp-admin/post.php:218 - include('/var/www/roomb...')
```

iandunn:
This is a duplicate of #311 

(The notices are slightly different, but essentially the same thing, and can all be fixed in a single issue)

cojennin:
Closing, fixed by #340

Issue #311, opened 2015-07-13
PHP Strict Standards:  Only variables should be assigned by reference


AramZS:
On activate, the following error occurs:

PHP Strict Standards:  Only variables should be assigned by reference in [...]wp-content/plugins/edit-flow/common/php/class-module.php on line 564, referer: [...]wordpress/wp-admin/update.php?action=install-plugin&plugin=edit-flow&_wpnonce=1ccd691aba

cojennin:
Closing, fixed by #340

PR #340, opened 2016-04-26
Cleanup PHP to comply with strict standards

cojennin:
Encountering the "only variables should be assigned by reference" and "only variables should be passed by reference" warnings when PHP strict standards are enabled. Cleanup instances in the code where these warnings are being triggered.

**fix for #311**
`$role = get_role( $role );`
Don't think we need the `&` if I'm reading http://php.net/manual/en/language.oop5.references.php correctly. Can someone with a bit more PHP expertise verify that?

**fix for #321**
`$default_status = array_shift( $custom_statuses );`
Looks like we need to pass a variable to `array_shift` instead of the return value of `$this->get_custom_statuses() `

**edit:** tagging issue correctly

### DIFF
--- a/tests/test-class-module.php
+++ b/tests/test-class-module.php
@@ -1,0 +1,51 @@
+<?php
+
+class WP_Test_Edit_Flow_Class_Module extends WP_UnitTestCase {
+	protected static $admin_user_id;
+
+	public static function wpSetUpBeforeClass( $factory ) {
+		self::$admin_user_id = $factory->user->create( array( 'role' => 'administrator' ) );
+	}
+
+	function _flush_roles() {
+		// we want to make sure we're testing against the db, not just in-memory data
+		// this will flush everything and reload it from the db
+		unset( $GLOBALS['wp_user_roles'] );
+		global $wp_roles;
+		if ( is_object( $wp_roles ) ) {
+			$wp_roles->_init();
+		}
+	}
+
+	function setUp() {
+		parent::setUp();
+		// keep track of users we create
+		$this->_flush_roles();
+
+	}
+	
+	function test_add_caps_to_role() {
+		$EditFlowModule = new EF_Module();
+
+		$usergroup_roles = array(
+			'administrator' => array( 'edit_usergroups' ),
+		);
+
+		foreach( $usergroup_roles as $role => $caps ) {
+			$EditFlowModule->add_caps_to_role( $role, $caps );
+		}
+
+		$user = new WP_User( self::$admin_user_id );
+
+		//Verify before flush
+		$this->assertTrue( $user->has_cap( 'edit_usergroups' ), 'User did not have role edit_usergroups' );
+
+		$this->_flush_roles();
+
+		$this->assertTrue( $user->has_cap( 'edit_usergroups' ), 'User did not have role edit_usergroups' );
+	}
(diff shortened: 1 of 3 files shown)


cojennin:
Added a unit test to make sure EF_Module `add_caps_to_role` correctly adds capabilities to a role.

tests/test-class-module.php
```diff
@@ -0,0 +1,47 @@
...
+		);
+
+		foreach( $usergroup_roles as $role => $caps ) {
+			$EditFlowModule->add_caps_to_role( $role, $caps );
+		}
+
+		$this->_flush_roles();
+
+		$user = new WP_User( self::$admin_user_id );
+		$this->assertTrue( $user->exists(), "Problem getting user" );
```
mjangda: I'm not sure if the user exists assertion is necessary in this test?

tests/test-class-module.php
```diff
@@ -0,0 +1,47 @@
...
+
+		$usergroup_roles = array(
+			'administrator' => array( 'edit_usergroups' ),
+		);
+
+		foreach( $usergroup_roles as $role => $caps ) {
+			$EditFlowModule->add_caps_to_role( $role, $caps );
+		}
+
+		$this->_flush_roles();
```
mjangda: We should probably check the caps before the role flush as well? (Or could split up into two tests.)

tests/test-class-module.php
```diff
@@ -0,0 +1,47 @@
...
+	protected static $admin_user_id;
+
+	public static function wpSetUpBeforeClass( $factory ) {
+		self::$admin_user_id = $factory->user->create( array( 'role' => 'administrator' ) );
+	}
+
+	function _flush_roles() {
+		// we want to make sure we're testing against the db, not just in-memory data
+		// this will flush everything and reload it from the db
+		unset($GLOBALS['wp_user_roles']);
```
mjangda: Minor: spaces around parentheses.

tests/test-class-module.php
```diff
@@ -0,0 +1,47 @@
...
+		global $wp_roles;
+		if ( is_object( $wp_roles ) )
```
mjangda: Minor: `{ }` around `if` block.

tests/test-class-module.php
cojennin: Yea, was wondering about that. In the `test_role_add_cap` function [capabilities.php](https://develop.svn.wordpress.org/tags/4.5/tests/phpunit/tests/user/capabilities.php) a call to `_flush_roles` appears after every call to `add_cap`.

Do you know if there's a reason for that? That's why I added that `_flush_roles` call in the first place

tests/test-class-module.php
cojennin: Was getting some issues in the CI tests on Travis where the user didn't seem to exist, so had it there for clarification. If the tests pass on these most recent changes I'll remove it.

cojennin:
@mjangda [added the cap check] before `_flush_roles`. I imagine it can't hurt, and it causes issues will fix as they arise. 

Seems like the issue I was seeing earlier with the user not being properly created is gone as well, so removed that assertion.

mjangda:
👍

cojennin:
merging

PR #343, opened 2016-04-29
Custom status timestamp fix

cojennin:
Working on fixing up custom status bugs. Taking a crack at reworking this method and adding as many tests as possible around the custom status functionality since it seems to be a persistent and pernicious issue.


cojennin:
@kraftbj one thing I noticed from the tests: there's a [check you added](https://github.com/Automattic/Edit-Flow/blame/master/modules/custom-status/custom-status.php#L1331) to see if $_POST is empty. Do you know why you added that in?

cojennin:
## Custom status timestamp bug
- In [wp_insert_post](https://core.trac.wordpress.org/browser/tags/4.5.1/src/wp-includes/post.php#L3182), if the post_date_gmt of the post being saved is empty or `'0000-00-00 00:00:00'` and the post status is not `'draft', 'pending' or 'auto-draft'`, the post_date_gmt field will be set. Custom statuses are not always equal to `'draft', 'pending' or 'auto-draft'`, so post's with custom statuses will have their post_date_gmt set
- Bug in the wild: #301, #268

## This patch fixes this issue by
- Checking to see if the post_status is 'future' or 'publish' and returning early without making any changes to post_date_gmt (fix for #268)
- If the incoming post data had an empty post_date_gmt field, keep the post_date_gmt field empty

## Tested by
[WP_Test_Edit_Flow_Custom_Status](https://github.com/Automattic/Edit-Flow/pull/343/files#diff-84e591ce27341a09991a958fab474ccdR3) - This tests combinations of different post_status with the post_date_gmt field set and unset.
[WP_Test_Edit_Flow_Custom_Status_Ajax](https://github.com/Automattic/Edit-Flow/pull/343/files#diff-df02cad6e7abc4e31f5cab9e94323a24R5) - This tests to make sure autosaving doesn't mess with the post_date_gmt field.

cojennin:
@mjangda @samglover @kel @kraftbj If you folks have some time it'd be great to get a review of this

kel:
@cojennin I'm not on the project any more and don't have access to test this on there anymore. Sorry.

cojennin:
@kel No worries, thanks for checking back in!

cojennin:
If no objections will merge shortly

cojennin:
Tests look good. Tried to cover as many use cases as possible with the tests (including autosaving) but if anyone notices a missed case let me know.

PR #345, opened 2016-05-10
Fix $post_int init

cojennin:
Doing some minor cleanup on @jeremyclarke's PR #328 per @mjangda's comments.

@jeremyclarke @mjangda do you have a minute to review?

### DIFF
--- a/common/php/class-module.php
+++ b/common/php/class-module.php
@@ -253,7 +253,7 @@ class EF_Module {
 	function get_current_post_type() {
 		global $post, $typenow, $pagenow, $current_screen;
 		//get_post() needs a variable
-		$post_int;
+		$post_int = '';
 		if( isset( $_REQUEST['post'] ) )
 			$post_int = (int)$_REQUEST['post'];
 
--- a/common/php/class-module.php
+++ b/common/php/class-module.php
@@ -253,27 +253,25 @@ class EF_Module {
 	function get_current_post_type() {
 		global $post, $typenow, $pagenow, $current_screen;
 		//get_post() needs a variable
-		$post_int = '';
-		if( isset( $_REQUEST['post'] ) )
-			$post_int = (int)$_REQUEST['post'];
+		$post_id = isset( $_REQUEST['post'] ) ? (int)$_REQUEST['post'] : false;
 
-		if ( $post && $post->post_type )
+		if ( $post && $post->post_type ) {
 			$post_type = $post->post_type;
-		elseif ( $typenow )
+		} elseif ( $typenow ) {
 			$post_type = $typenow;
-		elseif ( $current_screen && isset( $current_screen->post_type ) )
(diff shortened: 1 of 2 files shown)


common/php/class-module.php
```diff
@@ -253,27 +253,25 @@ function enqueue_datepicker_resources() {
 	function get_current_post_type() {
 		global $post, $typenow, $pagenow, $current_screen;
 		//get_post() needs a variable
-		$post_int;
-		if( isset( $_REQUEST['post'] ) )
-			$post_int = (int)$_REQUEST['post'];
+		$post_int = isset( $_REQUEST['post'] ) ? (int)$_REQUEST['post'] : false;
```
mjangda: Minor: it'd be nice to rename the variable to `$post_id`.

tests/test-class-module.php
```diff
@@ -44,6 +44,44 @@ function test_add_caps_to_role() {
...
+
+		unset( $_REQUEST['post_type'] );
+
+		set_current_screen( 'edit.php' );		
+
+		$this->assertEquals( 'post', $EditFlowModule->get_current_post_type() );
+
+		_unregister_post_type( 'content' );
+
+		set_current_screen( 'front' );
```
mjangda: These should probably all be separate tests?

tests/test-class-module.php
```diff
@@ -1,10 +1,13 @@
 <?php
 
 class WP_Test_Edit_Flow_Class_Module extends WP_UnitTestCase {
+
 	protected static $admin_user_id;
+	protected static $EditFlowModule;
 
 	public static function wpSetUpBeforeClass( $factory ) {
 		self::$admin_user_id = $factory->user->create( array( 'role' => 'administrator' ) );
+		self::$EditFlowModule = new EF_Module();
```
mjangda: We should probably instantiate a new module for each test (i.e. in `setUp`) and de-init in `tearDown` (i.e. `self::$EditFlowModule = null`).

mjangda:
Nice! One more note and then 🚢

cojennin:
@mjangda made the tweaks. Split the tests and did some cleanup

tests/test-class-module.php
cojennin: Is `self::$EditFlowModule = null` necessary in the tearDown? (Just curious, don't know much about PHP GC)

mjangda: Not strictly necessary, I guess, but it's a good idea to clean up after ourselves (we'd end up with reference after the last test is run).

cojennin: Gotcha, added.

mjangda:
🚢 🚢 🚢  (once Travis finishes)

cojennin:
Tests look good, merging.

Thanks @jeremyclarke for kicking this off with #328

jeremyclarke:
Thanks guys and sorry I didn't get to it!

PR #347, opened 2016-05-12
Add SVN deloyment script

cojennin:
Adding a script to help with SVN deployments based off of the that [exists in Jetpack](https://github.com/Automattic/jetpack/blob/master/tools/deploy-to-svn.sh).


cojennin:
Merging this to help with pushing to SVN.

PR #349, opened 2016-05-12
[WIP] Emulate get_sample_permalink_html

cojennin:
`get_sample_permalink_html` checks the status of a post in order to generate the html needed to support editing the permalink. To circumvent this we'll leverage the get_sample_permalink_html filter to rework the html after get_sample_permalink_html has run.

### DIFF
--- a/modules/custom-status/custom-status.php
+++ b/modules/custom-status/custom-status.php
@@ -1444,6 +1444,13 @@ class EF_Custom_Status extends EF_Module {
 	 * manipulating the slug. Critical for cases like editing the sample permalink on
 	 * hierarchical post types.
 	 * @since 0.8.2
+	 *
+	 * @param string  $permalink Sample permalink
+	 * @param int 	  $post_id 	 Post ID
+	 * @param string  $title 	 Post title
+	 * @param string  $name 	 Post name (slug)
+	 * @param WP_Post $post 	 Post object
+	 * @return string $link Direct link to complete the action
 	 */
 	public function fix_get_sample_permalink( $permalink, $post_id, $title, $name, $post ) {
 		//Should we be doing anything at all?
@@ -1494,13 +1501,20 @@ class EF_Custom_Status extends EF_Module {
 	/**
 	 * Hack to work around post status check in get_sample_permalink_html
 	 * 
-	 * @since 0.8.2
 	 * 
 	 * The get_sample_permalink_html checks the status of the post and if it's
 	 * a draft generates a certain permalink structure.
 	 * We need to do the same work it's doing for custom statuses in order
 	 * to support this link
 	 * @see https://core.trac.wordpress.org/browser/tags/4.5.2/src/wp-admin/includes/post.php#L1296
+	 *
+	 * @since 0.8.2
+	 * 
+	 * @param string  $return    Sample permalink HTML markup
+	 * @param int 	  $post_id   Post ID
+	 * @param string  $new_title New sample permalink title
+	 * @param string  $new_slug  New sample permalink kslug
+	 * @param WP_Post $post 	 Post object
 	 */
 	function fix_get_sample_permalink_html( $return, $post_id, $new_title, $new_slug, $post ) {
 		$status_slugs = wp_list_pluck( $this->get_custom_statuses(), 'slug' );
--- a/modules/custom-status/custom-status.php
+++ b/modules/custom-status/custom-status.php
@@ -103,12 +103,12 @@ class EF_Custom_Status extends EF_Module {
 		add_action( 'admin_init', array( $this, 'check_timestamp_on_publish' ) );
 		add_filter( 'wp_insert_post_data', array( $this, 'fix_custom_status_timestamp' ), 10, 2 );
 		add_action( 'wp_insert_post', array( $this, 'fix_post_name' ), 10, 2 );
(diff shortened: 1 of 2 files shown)


cojennin:
Without this fix, the link in the screenshot below returns a 404

screen shot 2016-05-12 at 3 37 29 pm

@philipjohn @mjangda do you have a minute to review? The tests used were pulled from https://develop.svn.wordpress.org/tags/4.5/tests/phpunit/tests/admin/includesPost.php.

This feels like a lot of work to tackle just this bug, so if anyone's got thoughts on how to minimize I'm all ears.

modules/custom-status/custom-status.php
```diff
@@ -1488,6 +1489,72 @@ public function fix_preview_link_part_two( $permalink, $post ) {
...
+			return $return;
+		}
+
+		$view_link = $this->get_preview_link( $post );
+		$preview_target = " target='wp-preview-{$post->ID}'";
+		list($permalink, $post_name) = get_sample_permalink( $post->ID, $new_title, $new_slug );
+
+		// Permalinks without a post/page name placeholder don't have anything to edit
+		if ( false === strpos( $permalink, '%postname%' ) && false === strpos( $permalink, '%pagename%' ) ) {
+			$modified_return = '<strong>' . __( 'Permalink:' ) . "</strong>\n";
```
cojennin: Wasn't sure if this was supposed to be a domain here?

modules/custom-status/custom-status.php
```diff
@@ -1501,6 +1568,7 @@ private function get_preview_link( $post ) {
 		} else if ( 'post' == $post->post_type ) {
 			$args = array(
 					'p'          => $post->ID,
+					'preview'	 => 'true'
```
cojennin: Unsure if this is needed for the other post_type's, but the unit tests based off of core require this. Maybe add a unit test for page or a custom post_type.

tests/test-edit-flow-custom-status.php
```diff
@@ -5,10 +5,13 @@ class WP_Test_Edit_Flow_Custom_Status extends WP_UnitTestCase {
 	protected static $admin_user_id;
 
 	public static function wpSetUpBeforeClass( $factory ) {
+		$EF_Custom_Status = new EF_Custom_Status();
```
cojennin: Thought we ran `init` on all our modules as a part of boostrapping the unit tests. But without this init the tests didn't work as expected (it appeared the custom statuses weren't registered).

cojennin:
Needs more work to handle pages

philipjohn:
@cojennin I'm unclear about the bug here, could you explain a bit more? I'm struggling to reproduce at the moment.

cojennin:
@philipjohn definitely! Not jazzed about the amount of churn in this change, so if there's an easier way to correct this (or I'm just doing something wrong) let me know!

Two issues around permalinks (that I've noticed) that this patch will fix below.

## Steps to reproduce
**Posts**
1. Create a new post
2. Save it with a status of "Pitch"
3. In the screenshot below, notice that the permalink shows the full link, but the "href" tag is missing "github-issue-349". So instead of linking to http://local.wordpress.dev/2016/05/25/github-issue-349/, it's linking to http://local.wordpress.dev/2016/05/25// (missing **github-issue-349**)

link-issue-1

**Hierarchical Post Types (demonstrated using "Pages")**
1. Create Page A and publish it
2. Create Page B and set the parent of Page B to Page A
3. In the screenshot below, when I go to set the parent of Page B to Page A, an autosave action occurs and the permalink for Page B is missing the hierarchical portion of the permalink (**expected, this happens without Edit Flow activated**)
4. Save Page B with a status of "Pitch"

After I save Page B with a status of "Pitch" and as a child of Page A, I expect the permalink to accurately reflect the hierarchy (note the slug in the "What I expect" screenshot contains "parent-page"). But that doesn't happen with Edit Flow activated.

**What I get**

link-issue-2


**What I expect** (notice that the permalink highlighted correctly reflects the hierarchy)

link-issue-3

philipjohn:
Very thorough, thanks :) Reproduced both and confirmed this fixes them, at least. Do we have unit tests for this particular feature? Is that feasible?

cojennin:
@philipjohn heh, no problem! 

Should be some tests for this patch in [test-edit-fow-custom-status.php](https://github.com/Automattic/Edit-Flow/pull/349/files#diff-84e591ce27341a09991a958fab474ccd). Do you think those are thorough enough? I cherry picked and modified the tests for `get_sample_permalink` found in [includesPost](https://develop.svn.wordpress.org/tags/4.4/tests/phpunit/tests/admin/includesPost.php) from the core WordPress PHPUnit tests.

philipjohn:
Duhh I should have seen those! Yeah, they give us what we need and we can always expand on those if we want to. The only thing I'd suggest is that we get some inline documentation so that we know what's going on :)

cojennin:
@philipjohn sure! There should be some explanation on top of `fix_get_sample_permalink` and `fix_get_sample_permalink`. Also have links to point back to the corresponding areas in trac that are at issue.

Cleaned up the comments to add `param` to both. Think that's enough? Can add some more explanation if you think it'll help.

philipjohn:
:+1: thanks @cojennin !

PR #350, opened 2016-05-12
Modify custom_status_timestamp to cover more statuses.

cojennin:
Previously was just covering 'publish' and 'future'. Should also cover 'private'. Rather than trying to account for these, if the status is not in this array return early.


cojennin:
Tests pass, merging

PR #353, opened 2016-06-07
Documenting edit_flow.php

cojennin:
The start of documentation for Edit Flow. This PR documents edit_flow.php Action's and Method's (there are no Filter's in edit_flow.php).


cojennin:
@philipjohn thoughts on documentation? Figured I'd start going class by class and documenting the action's, filter's and useful method's in that class. 

For the moment I thought it'd be easier to get contributors to contribute documentation if it was kept in Github instead of on editflow.org. 

Thoughts?

philipjohn:
Great thinking, having this stuff in Git will definitely help!

To make this easier in the future, we should follow the [inline documentation standards](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/) for these. That way we can piggy back on the code that generates the [WP Code Reference](https://developer.wordpress.org/reference/) (or something similar) to automate documentation on EditFlow.org.

How does that sound?

cojennin:
Sounds great! 

Is the tool used for generating the WP Code Reference [phpDocumenter](https://www.phpdoc.org/)? Can just go back through this PR and correctly document those functions using the WP doc standards and then have that file auto-generate.

philipjohn:
> Is the tool used for generating the WP Code Reference phpDocumenter?

It looks like partly, yes. [WP-Parser](https://github.com/WordPress/phpdoc-parser) is what makes all the magic happen. Essentially, I think (I'm guessing here) it takes the inline documentation (written to the WP standards) and parses that into pretty docs.

My hope is we can just point that at the Edit Flow source, and commit those auto-generated docs somewhere.

philipjohn:
Oh, more information on the [make.wordpress.org page about 'devhub'](https://make.wordpress.org/docs/handbook/developer-resources/devhub/).

cojennin:
@philipjohn just getting back around to this

The WP CLI command looks like it can [generate some JSON](https://github.com/WordPress/phpdoc-parser/blob/master/lib/class-command.php#L20) that can later be [ingested as WordPress posts](https://github.com/WordPress/phpdoc-parser/blob/master/lib/class-command.php#L44).

Not 100% up with how VIP works, but would it be possible to import that JSON into http://editflow.org/?

cojennin:
@philipjohn the [phodoc.json](https://github.com/Automattic/Edit-Flow/blob/0167afd48699cdb2e0508ba977b5dbb220934f14/phpdoc.json) file that's part of this commit is what was generated when I ran [WP-Parser](https://github.com/WordPress/phpdoc-parser) on the Edit Flow codebase from my machine. Seems like it might work, just need some place to stash the documentation that's accessible. Any thoughts on where to put the docs? Hoping we could store them on editflow.org

cojennin:
@philipjohn let me know if you have any ideas for where to throw these docs! May try and stage a `docs.editflow.org` WordPress site (though not sure who owns the domain...) and upload them there to see how it would work.

PR #356, opened 2016-07-26
Remove PHP4 constructor from Screen Options class

mjangda:
For better PHP7+ compat.


cojennin:
Looks good! Think that's the last spot using a PHP4 constructor? Test failure is unrelated, should be passing again as of  6cd9632

Issue #344, opened 2016-05-01
v0.8.1 breaks post previews


jb510:
Not sure why yet, but post previews are failing on v0.8.1. 

I rolled back to 0.7.6, still failing, rolled up to 0.8.0 and they started working 😕 

I manually installed 0.8.1, again broken, went back to 0.8.0 for now and it's working fine 😖 

This is the primary install on a multisite (just a couple sub-sites).  I've recreated the issue on staging site as well.  We use CPTs and PFs, but this is just a standard post.

cojennin:
Having some trouble recreating this on my setup (Edit Flow v0.8.1, WordPress v4.5). 

Quick questions:
What version of WordPress are you using?
What kind of permalink structure do you have?
How are you previewing it? (hitting the preview button on the Post screen, hitting the preview link from the page listing all posts, etc)

Something quick to try:
After you install Edit Flow v0.8.1, go to the "Permalinks" settings page and click the "Save Changes" button. See if that helps with previewing.

jb510:
Thanks for trying to reproduce.

WP Version: 4.5 and 4.5.1
Permalinks: Day and name /2016/05/02/sample-post/
PHP: 5.6 (WP Engine)
Steps: Same behavior from edit post page (preview on post screen) as post list (hover, quick links)

I deactivated all plugins and it's still occurring...  TwentySixteen, still occurring.

I can debug line by line locally later.

cojennin:
Following up for some more info:

1. Just to make sure, when you say "failing" you mean they're throwing a 404? Or some other kind of failure
2. When you deactivate Edit Flow previewing starts working again?
3. When you activate Edit Flow but no other plugins previewing stops working?

jb510:
1. yes, WP 404
2. yes, deactivated or running 0.8.0 preview works, running 0.8.1 previews do not
3. yes, with _only_ editflow 0.8.1 active and running TwentySixteen, previews go to a 404

It would be great to get confirmation, but this isn't urgent. I'll debug later this week.

cojennin:
Hrm. Yea, having trouble trying to get this to reproduce. Haven't seen other reported instances of this issue, but if I do I'll tag them here.

Let me know what you find when you have some time to debug.

cojennin:
@jb510 did you have some time to dig into what might be causing this?

jb510:
not yet.

cojennin:
[...]

PR #357, opened 2016-07-27
Build home_url() previews with trailing slash

jeremyfelt:
`home_url()` will only add a trailing slash to a site's home URL if a path is passed as an argument.

If no path is passed, a page preview could be something like: `http://wp.wsu.dev?page_id=1`

And it would work fine in most cases.

In a subdirectory multisite configuration, a page preview could be something like: `http://wp.wsu.dev/site?page_id=1`

This could be processed by WordPress as a lookup for a page named "site" rather than a site with that URL because no trailing slash exists.

When adding query args to `home_url()`, we should tell it to use a slash to avoid confusion.

See `_get_page_link()` for how core builds these in a way where a slash is always added.

See also #344, which I believe this fixes.

### DIFF
--- a/modules/custom-status/custom-status.php
+++ b/modules/custom-status/custom-status.php
@@ -1605,7 +1605,7 @@ class EF_Custom_Status extends EF_Module {
 		}
 
 		$args['preview_id'] = $post->ID;
-		return add_query_arg( $args, home_url() );
+		return add_query_arg( $args, home_url( '/' ) );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


cojennin:
Awesome! Makes sense. Can you [add a test](https://github.com/Automattic/Edit-Flow/blob/master/tests/test-edit-flow-custom-status.php) to make sure this function returns the preview link correctly for both single and multisite installs? Tests should be passing again with [this most recent commit](https://github.com/Automattic/Edit-Flow/commit/6cd9632de634ee113b6dfff91c307b21a3df0f11).

cojennin:
Just following up. Think you'll have some time to write a test for this patch?

jeremyfelt:
Yep, I spaced on this a couple times. I'll have some time to add tests soon. :)

cojennin:
@jeremyfelt sounds good!

rinatkhaziev:
Marked for 0.8.3 release

PR #359, opened 2016-08-25
Added check for empty author in notifications

petenelson:
When creating a new page via WP-CLI using `wp post create` and not specifying the author, a PHP notice is occurring.

```
vagrant@vvv:/srv/www/wordpress-default/htdocs/wp-content$ wp post create post-content.html --post_type=page --post_title="New Page"
PHP Notice:  Trying to get property of non-object in /srv/www/wordpress-default/htdocs/wp-content/plugins/edit-flow/modules/notifications/notifications.php on line 567
```

This patch checks for an empty `$post_author` variable before trying to access the properties in the variable.

### DIFF
--- a/modules/notifications/notifications.php
+++ b/modules/notifications/notifications.php
@@ -563,8 +563,8 @@ jQuery(document).ready(function($) {
 			
 			$body .= sprintf( __( '== %s Details ==', 'edit-flow' ), $post_type ) . "\r\n";
 			$body .= sprintf( __( 'Title: %s', 'edit-flow' ), $post_title ) . "\r\n";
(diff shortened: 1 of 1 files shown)


cojennin:
Funky. Looks good!

If you could shift [this comment](https://github.com/Automattic/Edit-Flow/pull/359/files#diff-59ee072dc425b09b6d748cb84634734bR566) back into place (above the `Author: %1$s (%2$s)` statement, inside the new `if` statement) I'll merge this

cojennin:
Awesome, thanks! Will merge now.

(Note to self: would definitely be useful to have some unit testing around notifications)

Issue #324, opened 2015-10-26
Custom statuses are not installed if tags exist already?


DrLightman:
I've installed the public 0.8 plugin and in some websites I couldn't see all the five predefined custom statuses:

pitch
assigned
in-progress
draft
pending

Checked the code and found this in \edit-flow\modules\custom-status\custom-status.php line 166:

```php
		// Okay, now add the default statuses to the db if they don't already exist 
		foreach( $default_terms as $term ) {
			if( !term_exists( $term['term'] ) )
				$this->add_custom_status( $term['term'], $term['args'] );
		}
```

But what if they already exist?

DrLightman:
The test is made on the term name, not slug. (localized)

DrLightman:
Also, there need to be a function to completley wipe all Edit-Data from a WP installation. I'm having an hard time to let the plugin reinstall module data.

cojennin:
> Maybe you had to use a taxonomy for these.

Yea, think we should be scoping [the call to `term_exists`](https://github.com/Automattic/Edit-Flow/blob/master/modules/custom-status/custom-status.php#L169) by passing the optional taxonomy parameter. Will work on a fix and test it.

> Also, there need to be a function to completley wipe all Edit-Flow from a WP installation. I'm having an hard time to let the plugin reinstall module data as if it were the first time.

Yea, would be good to investigate the tear down procedures for Edit Flow when it's uninstalled. Should open that as a separate issue.

cojennin:
Hrm, looking at this again, seems like the term_exists check is in place for [the reason highlighted in this comment](https://github.com/Automattic/Edit-Flow/blob/master/modules/custom-status/custom-status.php#L769). 

Could be an odd looking slug is created if the term already exists (even if under a different taxonomy). Will look into it.

cojennin:
Going to close this. Will work on documenting this behavior (and the reason for it).

Issue #82, opened 2012-04-21
Reduce custom status conflicts with existing taxonomy terms


danielbachhuber:
By the nature of design, custom statuses are stored as terms in a 'post_status' taxonomy. This allows us to persist the custom status over time, as well as store extra information associated with the status.

As of v0.7.1, the taxonomy terms aren't prefixed because there's a twenty character limit for term slugs. This can have the nasty side-effect of colliding with other taxonomy terms, however ([example](http://wordpress.org/support/topic/plugin-edit-flow-problems-with-custom-statuses)). Basically, if there's a post_tag term called 'draft' or similar, it's not possible to re-register 'draft' as your custom status.

To get around this, we'd need to start prefixing our terms to reduce the likelihood of conflict. We should discuss this at depth though, because a lot of code would have to change.

PR #362, opened 2016-09-04
Use taxonomy when checking for term existence

shadyvb:


cojennin:
@shadyvb did you verify that this patch does not cause a regression based on [this comment](https://github.com/Automattic/Edit-Flow/pull/362/files#diff-228291de10e40133681b1ff8414a9a54R768)?

Could you provide a a test and before/after of the expected behavior? A test would help with verifying functionality and a before/after would help with setting expectations for new behavior.

First thoughts: seems like it might cause some headache if folks have to deal with automatically created slugs they don't have control over (and they aren't able to predict what they slug will end up looking like)

cojennin:
Looked into this a bit more. Looks like it was discussed in #82 but closed as wontfix. Also mentioned in the [support forum](https://wordpress.org/support/topic/plugin-edit-flow-problems-with-custom-statuses/). Still not sure why these checks weren't originally scoped to a taxonomy. The comments around this code (as I mentioned in #324) are why I'm hesitant to make this change.

shadyvb:
@cojennin I'm having hard time understanding why this might not be needed, went through all the comments and references you posted, but still can't think of a solid reason why not to fix it.

I don't think the support forum comment about the missing posts assigned to a `draft` tag is related to the proposed fix, that `term_exist` call worked around the issue of colliding terms, but did not really fix it. ie: It deprives the user from the privilege that WordPress already provides, having similar terms under different taxonomies, and IMHO for no real reason, specially with a four line fix.

The patch does not cause a regression in my tests, and if you have a reason to think there might be in any scenario, let me know and I'll be happy to help write some tests to eliminate any concerns. I just can't think of any related to this specific change ( not the support forum ticket ).

cojennin:
> I'm having hard time understanding why this might not be needed, went through all the comments and references you posted, but still can't think of a solid reason why not to fix it.

Same, which is why I'm confused. Seems strange to add the `term_exists` check but not include the taxonomy. Would think excluding it would be done for a specific reason. But maybe that's just overthinking it.

I'd like to get a minor version of Edit Flow out tomorrow or Friday with a few bug fixes, localization improvements. After that I'll merge this in.

Does seem strange that it was never brought up as a solution to #82 (I wish I remember why it wasn't brought up as an easy fix...)

cojennin:
Alrighty, Edit Flow v0.8.2 is out so going to merge this one in! Thanks for the patch!

PR #367, opened 2016-09-16
Restyle `configure-edit-flow-module` buttons.

cojennin:
The `configure-edit-flow-module` buttons on the Edit Flow settings
screen had a blue-tint applied to the text and button styling by
WordPress core. Adding two style changes to bring back the previous look
and feel.

### DIFF
--- a/modules/settings/lib/settings.css
+++ b/modules/settings/lib/settings.css
@@ -87,7 +87,9 @@
 	background: #DDD;
 	background: -webkit-gradient(linear, left top, left bottom, from(#BBB), to(#888));
 	background: -moz-linear-gradient(top,  #BBB,  #888);
+	box-shadow: none;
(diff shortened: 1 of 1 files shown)


cojennin:
Last minute style tweak before 0.8.2. Two line CSS change, going to merge.

jerclarke:
This might be a weird question, but what effect did this have? This seems to be the ticket related to this entry in the 0.8.3 changelog:

> - UI Improvement: Made primary buttons on Settings screen consistent with WordPress UI. Props cojennin.

But AFAICT the buttons are still super weird looking (as they have been for years):

ef-buttons-greenblue

That's current `master` branch running on 4.9.8

jerclarke:
FWIW I feel like the custom styles should just be disabled entirely. Let them be blue and avoid this problem coming back up in the future.

jerclarke:
Ah ok, I see what the change was, it's the "Configure buttons" which are "primary" but grey. Got it. 

The green ones having that same blue shadow has been driving me nuts forever.

PR #370, opened 2016-09-29
Russian translation deep update

achumakov:
An up-to-date Russian translation. Fully sync against current version. Supercedes previous pull request.

If POedit issues 3 plural warnings regarding both original .pot file and the translation--it's OK, it's a bug in POedit itself.


cojennin:
@achumakov thanks for the translation! Going to merge 👍 

If you're interested and you've got some time, it would be great to see this translation work up on https://translate.wordpress.org/locale/ru/default/wp-plugins/edit-flow

achumakov:
I took a look at that...

It looks like there's almost no activity—4 people suggested as few as 10 
strings during last year.

Also, the core of translation is not very good as it was made almost 
without any reference to both WP terminology guidelines (e.g. a widget 
which should be just transliterated is translated as 'attached module', 
etc), and Russian traditions of editorial workflow.

Plus some strings are too long for their on-screen real estate.

If we keep it as is, there is the risk of outdated version heading 
forward to replace new translation...


... I would better upload new .po there, for any newcomers to start from 
current point.


Regarding Readme, it's completely untranslated for now, and I could to 
the job soon if there's noboby else ;) I didn't know it is needed.


Cheers,

Alex


30.09.2016 0:05, Connor Jennings пишет:

achumakov:
BTW, I'm unsure where to put localized readme — could you suggest?


30.09.2016 0:37, Alexey Chumakov пишет:

PR #371, opened 2016-09-29
i18n dates for calendar and story budget

achumakov:
As suggested to me- separate pull request from Russian translation update.


cojennin:
@achumakov changes look great! Very happy to see these internationalization improvements 😄  Really helps make Edit Flow accessible to all

PR #374, opened 2016-10-01
Days of week and months are localized properly now

achumakov:
Used date_i18n not WP_locale as either the latter behaves strange or I just didn't manage to understand that.

Anyway, with date_i18n both days and months work as they should.

### DIFF
--- a/modules/calendar/calendar.php
+++ b/modules/calendar/calendar.php
@@ -1158,7 +1158,7 @@ class EF_Calendar extends EF_Module {
 		$html = '';
 		foreach( $dates as $date ) {
 			$html .= '<th class="column-heading" >';
-			$html .= esc_html( date('l', strtotime( $date ) ) );
+			$html .= esc_html( date_i18n('l', strtotime( $date ) ) );
 			$html .= '</th>';
 		}
 		
--- a/modules/calendar/calendar.php
+++ b/modules/calendar/calendar.php
@@ -668,7 +668,7 @@ class EF_Calendar extends EF_Module {
 				<tbody>
 				
 				<?php
-				$current_month = date( 'F', strtotime( $filters['start_date'] ) );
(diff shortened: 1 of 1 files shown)


modules/calendar/calendar.php
```diff
@@ -690,10 +690,10 @@ function view_calendar() {
 				<?php if ( $split_month ): ?>
 				<tr class="month-marker">
 					<?php foreach( $week_dates as $key => $week_single_date ) {
-						if ( date( 'F', strtotime( $week_single_date ) ) != $split_month && date( 'F', strtotime( "+1 day", strtotime( $week_single_date ) ) ) == $split_month ) {
-							$previous_month = date( 'F', strtotime( $week_single_date ) );
+						if ( date_i18n( 'F', strtotime( $week_single_date ) ) != $split_month && date_i18n( 'F', strtotime( "+1 day", strtotime( $week_single_date ) ) ) == $split_month ) {
```
cojennin: @achumakov Using `date_i18n` in these `if` statements won't cause any issues, right? Not very familiar with translations, so just wanted to double check.

cojennin:
PR looks good! Just had a quick question about the the `date_i18n` usage, other than that everything seems to be working great!

Is there a way to test any of these functions to make sure they're localizing properly? Not familiar with PHPUnit and localizing, but might be good to add one in here!

achumakov:
I've checked them in various locales, and analyzed the code. They work great _if_ all of them in the code block are either date_i18n or just date.

The code is just fetching month names of posts in the range, and if they differ, switches to a 'split month' output.

While all the checks are consistently calling the same function, all comparisons work with any locale.

cojennin:
Gotcha. Alright, checked it out with a few different locales as well and it looks good, going to merge!

PR #380, opened 2016-10-18
Fix get_beginning_of_week

FewKinG:
The method get_beginning_of_week has some flaws which I addressed in this commit.
1. The comments make no sense (mixing up end of week and beginning of week)
2. When the passed $date is not a start of the week and the passed number $week is greater than 1, the method will behave incorrectly, because the difference between $day_of_week and $start_of_week is then subtracted multiple times. In the current calendar implementation this seems to have no negative effect, because the passed $date will always be a start of a week. But I thought it would be proper to fix it anyway.
3. When, by means of the given offset of $week, a DST switch introducing additional time occurs during the period between $date and the final date, the method will also behave incorrectly (the same date(s) may occur multiple times in the calendar causing misalignment of weekdays and possible confusion)

### DIFF
--- a/modules/calendar/calendar.php
+++ b/modules/calendar/calendar.php
@@ -1314,8 +1314,8 @@ class EF_Calendar extends EF_Module {
 		$end_of_week = get_option( 'start_of_week' ) - 1;
 		$day_of_week = date( 'w', $date );
 		$date += (( $end_of_week - $day_of_week + 7 ) % 7) * 60 * 60 * 24;
-		$additional = 3600 * 24 * 7 * ( $week - 1 );
-		$formatted_end_of_week = date( $format, $date + $additional );
+		$date = strtotime ( '+' . ( $week - 1 ) . ' week', $date ) ;
+		$formatted_end_of_week = date( $format, $date );
 		return $formatted_end_of_week;
 		
 	}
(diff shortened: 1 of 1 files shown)


FewKinG:
Concerning the DST problem, I replaced the calculation of the week offset from which was done by adding a fixed amount of seconds with the use of strtotime like this:
`$date = strtotime ( '+' . ( $week - 1 ) . ' week', $date ) ;`

I'm not all too familiar with PHP but I hope this will be fine. At least this seems to respect DST changes.

FewKinG:
I added another commit, also fixing the same error in the corresponding method get_ending_of_week.

In this case the error resulted in the database query selecting a too short week range for weeks after the DST change whereby all sundays after the change appeared empty in the calendar.

cojennin:
> When, by means of the given offset of $week, a DST switch introducing additional time occurs during the period between $date and the final date, the method will also behave incorrectly (the same date(s) may occur multiple times in the calendar causing misalignment of weekdays and possible confusion)

jon-stewart-mind-blown

Crazy find! Given how gnarly this method is (and how critical it is to making sure the calendar renders correctly) can you add a [test-edit-flow-calendar.php](https://github.com/Automattic/Edit-Flow/tree/master/tests) file with a few tests that make sure it displays correctly in normal situations as well as funky ones (i.e., DST change in between start and end of week). Also, would hopefully test a few international DateTime situations as well (if you know of any interesting one's to test for)!

Will review after that test is in and then work on merging. Thanks!

rinatkhaziev:
This looks good to me, however, I'm hesitant to include it in 0.8.3 without tests. I'm tentatively marking this for 0.8.3. @sboisvert care you chime in?

@FewKinG will you be open to adding tests? I realize that it's been way too long since the PR, and priorities most likely have shifted for you :)

PR #382, opened 2016-11-03
Check if $post_count object has property before trying to use it

jvolfson:
In the dashboard, if the $post_count object has no results for a custom status (for instance, if you create a status called 'in progress', but have not yet assigned that status to a post), line 166 will generate an error:

Notice:  Undefined property: stdClass::$in-progress in /var/www/html/web/app/plugins/edit-flow/modules/dashboard/dashboard.php on line 166

Added a check to see if the object has the property before attempting to use it.


rinatkhaziev:
Per discussion in #383 this doesn't seem to be an issue anymore

PR #386, opened 2016-11-20
A new Module for Edit Flow: "My Module"

michaelscheurer:
Hi! 

Edit Flow is a great Wordpress plugin. I wrote a new module for Edit Flow. With them you can easy create a new module template within one click. It should attract new developers.

"My Module" creates the basic folder structure and needed files (with runable content). A new created custom module is directly installed und runs on Edit Flow.

Improvements will be necessary concerning the user input and file permissions on the server. But the plugin works!

Would be glad if you can check the module and include it to Edit Flow.

I look forward to get feedback from you.

Thank you.

Michael


rinatkhaziev:
Thanks for your work on this one! 

Unfortunately, we won't be able to merge it - the approach won't work on any setup where write operations with file system are prohibited. 

I opened a new issue regarding extending Edit Flow functionality with custom modules #462

PR #396, opened 2017-06-13
Fix ef_pre_insert_editorial_comment filter

sudar:
The `apply_filters` function is getting called wrongly. (ie) The return value of the function is not getting stored anywhere, esentially rendering the filter useless.

This PR fixes that and also passes `$_POST` as an additional parameter to the filter.

### DIFF
--- a/modules/editorial-comments/editorial-comments.php
+++ b/modules/editorial-comments/editorial-comments.php
@@ -326,7 +326,7 @@ class EF_Editorial_Comments extends EF_Module
 			    'comment_approved' => self::comment_type,
 			);
 
-			apply_filters( 'ef_pre_insert_editorial_comment', $data );
+			$data = apply_filters( 'ef_pre_insert_editorial_comment', $data );
 
 			// Insert Comment
 			$comment_id = wp_insert_comment($data);


philipjohn:
Thanks for this @sudar ! What's the reason for including `$_POST`?

sudar:
@philipjohn 

I was trying to edit the values of the comment before it was saved. That's why I needed the `$_POST`.

One example scenario is to allow over ride the `wp_kses_post` call for comment content with `wp_kses` to allow iframes.

I can still read the user inputed value directly from `$_POST` inside my callback, but it looked more cleaner to pass it as a parameter to the filter itself.

sudar:
@philipjohn 

If you don't want to add the extra parameter that is also fine with me. I can send another PR without it.

philipjohn:
> If you don't want to add the extra parameter that is also fine with me. I can send another PR without it.

Yeah, let's do that for now to get the `$data` fix in, and we can have a wider discussion about the value of including `$_POST` in a new PR. Thanks!

sudar:
@philipjohn 

I have modified the PR to only fix the filter.

Once this get's in, I will create another with `$_POST` and we can discuss about it there.

sudar:
@philipjohn 

Thanks.

PR #401, opened 2017-10-25
Added user_group object filter for better flexibility

tcrsavage:
This was required for one of our projects to allow filtering of user_ids on the object.

### DIFF
--- a/modules/user-groups/user-groups.php
+++ b/modules/user-groups/user-groups.php
@@ -844,6 +844,9 @@ class EF_User_Groups extends EF_Module {
 				$usergroup->$key = $value;
 			}
 		}
+
+		$usergroup = apply_filters( 'ef_usergroup_object', $usergroup );
+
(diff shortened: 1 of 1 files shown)


mjangda:
Can you share one or two use cases for this?

We would also want to make sure user group objects aren't accessed any other way to ensure consistency everywhere.

tcrsavage:
My use case was quite non-standard. I needed to add some multisite capabilities to notifications and that meant being able to override user ids subscribed to a list because of site member checks. In my case a filter just for subscribed user IDs would be sufficient.  A full object filter is a lot more flexible, if a little more dangerous for developers.

get_usergroup_by looks to be the common function as it's responsible for decoding the description data so anything dealing with groups outside of this will just be using it as a standard WP_Term.

rinatkhaziev:
Marking as 0.8.3

PR #402, opened 2017-10-31
Update Header Requirements

floSoX:
Hi!

Thanks you for your great plugin.

I propose an update to comply with this: https://wordpress.org/support/topic/error-on-install-does-not-have-a-valid-header/

For your information, it's discussed here: 
https://developer.wordpress.org/plugins/the-basics/header-requirements/

Regards,


philipjohn:
It looks like the issue was actually the nested trunk directory as pointed out in that thread. I've just removed that directory, which should fix the issue. Thanks!

Issue #377, opened 2016-10-01
i18n: Posts view / post properties inline block: Published status is unlocalized


achumakov:
(just putting it here to remember; will try to fix myself)

'Published' status is in English (only if you open inline properties dialog, with a drop-down to select a status).

All the others are localized properly.

raduconst:
Update: The PR makes the status goes from this:

<img width="213" alt="screenshot at jan 22 09-32-03" src="https://user-images.githubusercontent.com/6512220/35209521-2084944a-ff57-11e7-90c8-f3a837cda684.png">

to this:

<img width="231" alt="screenshot at jan 22 09-31-36" src="https://user-images.githubusercontent.com/6512220/35209527-24e7e92e-ff57-11e7-8e11-7586a8c5fe12.png">

PR #404, opened 2018-01-17
Status localize

raduconst:
This PR is intended to fix the issue #377 . Also, when checking this out I could see that the status dropdown from the right side of the page when adding/editing a post was not localized either. The options were all in English even if the language was set to French, for example.

### DIFF
--- a/common/php/class-module.php
+++ b/common/php/class-module.php
@@ -219,7 +219,7 @@ class EF_Module {
 		} else if ( array_key_exists( $status, $builtin_stati ) ) {
 			$status_friendly_name = $builtin_stati[$status];
 		}
-		return $status_friendly_name;
+		return esc_html__( $status_friendly_name, 'edit-flow');
 	}
 	
 	/**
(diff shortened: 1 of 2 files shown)


common/php/class-module.php
```diff
@@ -219,7 +219,7 @@ function get_post_status_friendly_name( $status ) {
 		} else if ( array_key_exists( $status, $builtin_stati ) ) {
 			$status_friendly_name = $builtin_stati[$status];
 		}
-		return $status_friendly_name;
+		return esc_html__( $status_friendly_name, 'edit-flow');
```
sboisvert: Good job adding the internationalization. Can you elaborate on why we'd want to escape here? It seems that we would be early escaping instead of escaping on output. There's a bit more information here: https://vip.wordpress.com/documentation/validating-sanitizing-escaping/#always-escape-late :)

modules/custom-status/custom-status.php
```diff
@@ -405,7 +405,7 @@ function post_admin_header() {
 			// Load the custom statuses
 			foreach( $custom_statuses as $status ) {
 				$all_statuses[] = array(
-					'name' => esc_js( $status->name ),
+					'name' => esc_js( esc_html__( $status->name, 'edit-flow') ),
```
sboisvert: Same as above regarding the translation being added properly but I question the esc_html__() and the esc_js(). I see it's currently present but I'm not sure I understand the purpose of the esc_js(). Here's a bit more information on https://vip.wordpress.com/2015/03/25/preventing-xss-in-javascript/

PR #405, opened 2018-01-18
Fix #243 - Offset post times to UTC+0 for Calendars

### LINKED ISSUE #243: iCalendar feed doesn't use gmt date for start/end

justnorris:
This fixes a problem where incorrect times were added to calendars because calendars expect a times to be in UTC+0 time zone. The problem described in #243 

Introduced `ics_format_time` static method that is formatting the time properly.

### DIFF
--- a/modules/calendar/calendar.php
+++ b/modules/calendar/calendar.php
@@ -8,12 +8,12 @@
 if ( !class_exists('EF_Calendar') ) {
 
 class EF_Calendar extends EF_Module {
-
+	
 	const usermeta_key_prefix = 'ef_calendar_';
 	const screen_id = 'dashboard_page_calendar';
-
+	
 	var $module;
-
+	
 	var $start_date = '';
 	var $current_week = 1;
 	var $total_weeks = 6; // default number of weeks to show per screen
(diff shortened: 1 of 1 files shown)


justnorris:
Removed spacing adjustments for a cleaner PR

modules/calendar/calendar.php
```diff
@@ -489,6 +489,33 @@ function do_ics_escaping( $text ) {
...
+
+		// Timestamp it
+		$timestamp = strtotime( $time_string );
+
+		if( ! $timestamp ) {
+			return false;
+		}
+
+		// Add GMT Offset
+		$timestamp += -1 * get_option('gmt_offset') * HOUR_IN_SECONDS;
```
sboisvert: I'm not sure I understand why we're doing += -1 *, could we not do -= instead?

modules/calendar/calendar.php
```diff
@@ -489,6 +489,33 @@ function do_ics_escaping( $text ) {
...
+
+		// Add manual offset
+		$timestamp += $offset_in_seconds;
+
+		// \T and \Z are escaped for literal T and Z characters
+		return date( 'Ymd\THis\Z', $timestamp );
```
sboisvert: This doesn't seem to be returning a timestamp like the function declaration says it is.

justnorris:
Cleaned up the method comment and the GMT calculation math. Everything should be dandy now :)

PR #407, opened 2018-01-19
Invert dashboard menu icon and remove old MP6 code

### LINKED ISSUE #238: Menu Icon doesn't work with MP6
image

Too dark, unfortunately.

justnorris:
This could still use some improvements (turning the icon into an SVG file), but this patch is at least makes the icon visible again. 

Also - I ran the images through ImageOptim on mac to reduce file size. 

Fixes #238


justnorris:
So I can merge it in?

PR #408, opened 2018-01-19
Avoid direct SQL Queries by using WP Date Query

### LINKED ISSUE #248: Update date range queries to use Date Query
No more direct SQL for us!

We'll need to bump our minimum required version to 3.6.

justnorris:
Fixes #248 
Possibly a breaking change: Removing method `post_where_week_range` from the `posts_where` filter.

### DIFF
--- a/modules/story-budget/story-budget.php
+++ b/modules/story-budget/story-budget.php
@@ -388,9 +388,19 @@ class EF_Story_Budget extends EF_Module {
 		// Filter for an end user to implement any of their own query args
      	$args = apply_filters( 'ef_story_budget_posts_query_args', $args );
 
-		add_filter( 'posts_where', array( $this, 'posts_where_range' ) );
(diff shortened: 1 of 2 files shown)


modules/story-budget/story-budget.php
```diff
@@ -388,9 +388,17 @@ function get_posts_for_term( $term, $args = null ) {
 		// Filter for an end user to implement any of their own query args
      	$args = apply_filters( 'ef_story_budget_posts_query_args', $args );
```
sboisvert: Would moving the filter to be right before the WP_Query be more useful? That way the new parameters we're adding an be filtered out.
I'd also make it clear somewhere in the PR's description that this could potentially be a breaking change. We should note it in the Release Notes when we release a new version.

justnorris: I missed that filter! I'm going to move it around.

PR #409, opened 2018-01-20
Story budget columns

raduconst:
The columns on this page were created through PHP file by wrapping each column into a different div. This way, the width couldn't be changed dynamically depending on the number of columns was set on the Screen Options side. Now, I changed this part to be a big list and change the width of the items through CSS by simply modifying the class of the big wrapper.

### DIFF
--- a/common/php/class-module.php
+++ b/common/php/class-module.php
@@ -219,7 +219,7 @@ class EF_Module {
 		} else if ( array_key_exists( $status, $builtin_stati ) ) {
 			$status_friendly_name = $builtin_stati[$status];
 		}
-		return esc_html__( $status_friendly_name, 'edit-flow');
+		return $status_friendly_name;
 	}
 	
 	/**
--- a/modules/story-budget/lib/story-budget.js
+++ b/modules/story-budget/lib/story-budget.js
@@ -27,8 +27,6 @@ jQuery(document).ready(function($) {
 			$postboxContainer.hasClass('columns-number-3') )
 			$postboxContainer.removeClass('columns-number-1 columns-number-2 columns-number-3');
 		$postboxContainer.addClass('columns-number-' + numColumns );
-
-		//jQuery(".postbox").css('width', (90 / numColumns) + '%' );
 	});
 	
 	jQuery('h2 a.change-date').click(function(){


raduconst:
This is another example that I should give some more details about the changes that I do on a commit.
The problem that I'm trying to fix through this PR is that the number of columns on the Story Budget page is not changing when using the option from Screen Options. For now, this is the behavior of the columns: https://cl.ly/1S1q310S0p2G
The problem is that the structure of the columns right now is not that good. The number of columns is not created dynamically based on what value you choose for the number of columns. For example, if you have one column set on _Screen Options_, the columns will look like this:
screenshot at jan 29 11-18-57
If you select to have two columns (from Screen Options), the columns will look like this:
screenshot at jan 29 11-20-07
The structure won't be changed. However, when the page is refreshed, the columns are created correctly:
screenshot at jan 29 11-21-22
The change I made will make things work like this: https://cl.ly/3l3K0n3H2D1s

rinatkhaziev:
Hi @raduconst, 

I reviewed the pull request and have to agree with @sboisvert, this PR only handles a subset of cases (for columns 1 to 3), however, in the module `$max_num_columns` is a filterable property that can contain any value. Thus, if a user sets the filter to any number bigger than 3 the proposed logic will break.

Here's my proposition:
1. Modify the javascript like that:
```javascript
// Change number of columns when choosing a new number from Screen Options
var columnsSwitch = $("input[name=ef_story_budget_screen_columns]")
columnsSwitch.click(function() {
  var numColumns = parseInt($(this).val());
  var classPrefix = 'columns-number-';
  $(".postbox-container").removeClass(function() {
    for (var index = 1, c = []; index <= columnsSwitch.length; index++) {
      c.push( classPrefix + index )
    }
    return c.join(' ');
  }).addClass(classPrefix + numColumns);
});
```
This will allow us to assign the correct class no matter max number of allowed columns.
2. Modify the static css to take advantage of flexbox:
```css
.postbox {
  display: block;
  margin: 0 auto 10px;
  justify-content: space-between;
}

.postbox-container {
  padding-right: 0;
  display: flex;
  flex-flow: row wrap;
}
```
3. Modify PHP to remove extra markup and a little inline style that controls columns width based on class:
```
<?php
  echo '<div class="postbox-container">';
  foreach( (array) $this->terms as $term ) {
    $this->print_term( $term );
  }

  echo '</div>';
?>
<style>
  <?php
    for ( $i = 1; $i <= $this->max_num_columns; ++$i ) {
      ?>
        .columns-number-<?php echo (int) $i; ?> .postbox {
          flex-basis: <?php echo  99 / $i ?>%;
        }
      <?php
    }
  ?>
</style>
</div>
```

4. Finally, make sure that it looks good on mobile, most likely we'll need to force columns always be 100% on mobile.

Let me know if you're open to make these changes.

raduconst:
Thanks @rinatkhaziev for checking this out. I'll make the changes today and update this PR to make sure it gets included in version 0.8.3.

rinatkhaziev:
Closing this in favor of #460

Issue #375, opened 2016-10-01
i18n: timesince function needs review


achumakov:
timesince() needs internationalization, as it currenly produces English-only output (M jS template; 'ago', etc.)

This is not really easy task as both PHP and Wordpress lack proper i18n mechanics for this task. 

Maybe, just fallback to datetime without any 'ago' attempts when current locale is not English.

raduconst:
I think that there's a function called _human_time_diff_ that can be used in this case. By changing the _post_modified_ case from the _term_column_default_ function and replacing everything with:
`return sprintf( esc_html__( '%s ago', 'edit-flow' ), human_time_diff( get_the_time( 'U', $post->ID ), current_time( 'timestamp' ) ) );`
the last column of the table will display automatically the right time difference without using the timesince function. I gave this a try on my side and everything works pretty well.

PR #410, opened 2018-01-22
Changed the way that the status on the calendar is made

raduconst:
All I did was to replace the _timesince_ function with the _human_time_diff_ function from WordPress. This way, the time difference is automatically created. This should do the trick for #375


rinatkhaziev:
Thanks, marking this for 0.8.3 release

PR #414, opened 2018-01-22
Fix #389 - Custom Statuses in Pages

### LINKED ISSUE #389: Custom Statuses not working on pages bulk edit
image

image

justnorris:
This fixes #389 and makes the `custom-status.js` translatable

### DIFF
--- a/modules/custom-status/custom-status.php
+++ b/modules/custom-status/custom-status.php
@@ -80,7 +80,6 @@ class EF_Custom_Status extends EF_Module {
 		// Load CSS and JS resources that we probably need
 		add_action( 'admin_enqueue_scripts', array( $this, 'action_admin_enqueue_scripts' ) );
 		add_action( 'admin_notices', array( $this, 'no_js_notice' ) );
-		add_action( 'admin_print_scripts', array( $this, 'post_admin_header' ) );
 
 		// Methods for handling the actions of creating, making default, and deleting post stati
 		add_action( 'admin_init', array( $this, 'handle_add_custom_status' ) );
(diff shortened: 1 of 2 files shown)


sboisvert:
Hi @justnorris!

This PR is a really good start. There are a few things I think could help make it perfect. With the current patch changing the add_action( 'admin_print_scripts') to wp_localize_script() could break Backwards compataility for some users. While wp_localize_script() is a much cleaner way of doing it and is more intuitive it's possible that some sites / other plugin authors (maybe even some localization plugins) are already handling this via the current 'post_admin_header' action hook. While if we were writing this code from scratch using wp_localize_script would be a better way of doing it, when maintaining community plugins we try to not break any potential customization users might of put it. Sometimes it's not possible but I think in this case we don't actually need to change it to fix the bug.

I also feel like the pull request message could be enriched. There are a few design decisions in your PR that we could expand on. For example there we're removing many JS variables (for example `custom_statuses`) and changing them to two arrays. While I agree that using 2 arrays and not polluting the global scope is good, I again worry that it could break some customization in user's themes and plugins. I don't think we need to make those changes to add the internationalization. It's again one of those that it would be better if we had started with that and the array is a better way of doing it (There might even be an edit flow object somewhere we could tack this on to). That being said, I think we owe it to the users to try to break the least amount of things possible and only improve the plugin. Therefore I wouldn't create the new `variables` variable. I think using an array for i18n is perfectly fine. I don't think we need to follow the same practice as is currently there since we're adding this as a new "feature".

There's also a change in behaviour that isn't referenced anywhere that I can see. On L366 there's a slight change in the behaviour. From my reading this is most likely to squash a potential PHP warning, but I'm not sure :). I believe the end result is exactly the same. Regardless it would be nice to mention it :).

Thanks!

justnorris:
Okay, I'll ensure backwards compatibility and split the i18n into a separate PR

> On L366 there's a slight change in the behaviour. From my reading this is most likely to squash a potential PHP warning, but I'm not sure :). I believe the end result is exactly the same. Regardless it would be nice to mention it :).

Yep. Because this pull is adding support for "Mass Edit" in Pages - sometimes there is no `$post` available, which is why I added the check there. What would be a good way to mention that?

sboisvert:
> Yep. Because this pull is adding support for "Mass Edit" in Pages - sometimes there is no $post available, which is why I added the check there. What would be a good way to mention that?

I'd put this in a separate PR that mentions the why of this change :)

Issue #417, opened 2018-01-23
Allow non-post type data to be injected into the calendar view


cklosowski:
I've been asked by a user of Post Promoter Pro to see if I can support adding social media shares to the Edit Flow calendar. After inspection it looks like all the data is very heavily relying on the `WP_Post` objects.

For integrations it would be ideal to allow a way to inject some data into the views. I have a little work in progress to get this going, and some sample code that seems to work.

Will submit a PR for the initial thought and hopefully can work forward to maybe get it merged in as a feature.

WPprodigy:
Closing this issue, the PR has been merged. Thanks @cklosowski

PR #418, opened 2018-01-23
Support adding non-post data into days of calendar #417

cklosowski:
Initial thoughts/proposal for #417

### DIFF
--- a/modules/calendar/calendar.php
+++ b/modules/calendar/calendar.php
@@ -754,10 +754,14 @@ class EF_Calendar extends EF_Module {
 						$this->hidden = 0;
 						if ( !empty( $week_posts[$week_single_date] ) ) {
 
-							$week_posts[$week_single_date] = apply_filters( 'ef_calendar_posts_for_week', $week_posts[$week_single_date] );
+							$week_posts[$week_single_date] = apply_filters( 'ef_calendar_posts_for_week', $week_posts[$week_single_date], $week_single_date );
 
-							foreach ( $week_posts[$week_single_date] as $num => $post ){ 
-								echo $this->generate_post_li_html( $post, $week_single_date, $num ); 
+							foreach ( $week_posts[$week_single_date] as $num => $post ){
+								if ( is_a( $post, 'WP_Post' ) ) {
+									echo $this->generate_post_li_html( $post, $week_single_date, $num );
+								} else {
+									echo apply_filters( 'ef_calendar_single_date_item', '', $num, $post );
+								}
 							} 
 
 						 } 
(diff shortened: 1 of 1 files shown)


cklosowski:
This would allow something like this to be used:
```
<?php

function ck_add_shares_calendar_data( $items, $week_single_date ) {

	$items[] = array(
		'type' => 'share',
		'content' => 'Testing',
	);

	return $items;
}
add_filter( 'ef_calendar_posts_for_week', 'ck_add_shares_calendar_data', 10, 2 );

function ck_add_shares_to_ef_calendar( $output, $num, $post ) {
	if ( $post['type'] === 'share' ) {
		$output = $post['content'];
	}
	return $output;
}
add_filter( 'ef_calendar_single_date_item', 'ck_add_shares_to_ef_calendar', 10, 3 );
```

Which results in the ability to add something similar to:
screen shot 2018-01-22 at 11 15 56 pm

this is devoid of styling at the moment, but is just a proposal to see if it would be considered.

justnorris:
Strict Yoda statements FTW.

Any thoughts on changing `pre_ef_***` to `ef_pre_***`? Would that make the naming/prefixes more consistent?

justnorris:
Alright, this looks like good to go. 

@sboisvert - any thoughts?

cklosowski:
Any updates here @justnorris @sboisvert ? Need anything else from me to get this :+1:'d and merged?

sboisvert:
@philipjohn this looks good from my end, if you're 👍 then let's merge it :)

rinatkhaziev:
@cklosowski @sboisvert @philipjohn 

I have two notes here: 

```
foreach ( $week_posts[$week_single_date] as $num => $post ){
  $output = apply_filters( 'ef_pre_calendar_single_date_item_html', false, $this, $num, $post, $week_single_date );
  if ( false === $output ) {
    $output = $this->generate_post_li_html( $post, $week_single_date, $num );
  }
  echo $output;
}
```
While this code is working, I feel like there's a room for improvement clarity-wise - `$output` is clearly intended to be a string, yet the initial value is false and the consequent check is also against a boolean. I propose to change the initial value for filter to an empty string `''` and then change the check to 
```
// either 
if ( ! $output ) {
// or 
if ( '' === $output ) {
```
This will help to better communicate the intent. 

Another suggestion regarding that piece is that it can be reduced to a ternary:

```echo $output ?:  $this->generate_post_li_html( $post, $week_single_date, $num );```

The other note is Coding Standards violations around spacing array `When referring to array items, only include a space around the index if it is a variable` and a missing space before the opening brace in foreach : https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#space-usage . I realize it might sound nit-picky and the plugin is ridden with CS inconsitencies, we should probably try avoiding adding up to the problem :)

PR #421, opened 2018-01-23
Fix the test errors

raduconst:
Changed the deprecated method for role init because the old method was deprecated. Also, the dev: precise to PHP 5.2 and 5.3 because these two versions are supported only on Precise.


david-binda:
Thanks for looking into this @raduconst !

Could you, please, elaborate on why the tests for PHP versions prior to 5.6 are being removed? I can see that upon switching to "precise", those were passing: https://travis-ci.org/Automattic/Edit-Flow/builds/332221603

The only outstanding issue is the PHP 7, which is producing following Fatal Error:

```
Class 'PHPUnit_Framework_TestCase' not found
```

Could you look into that one as well?

raduconst:
@david-binda I removed the old versions of PHP from the tests based on the discussion we had here: https://a8c.slack.com/archives/C18HWREE9/p1516805841000053
It's not a final decision or anything like that but we're trying to move and improve the test environment.

I made a change on the travis.yml file right now, trying to test everything using the latest version of WP.

I'd love to hear some of your thoughts about all these changes.

.travis.yml
```diff
@@ -14,34 +18,46 @@ env:
...
+      env: WP_VERSION=4.8 WP_MULTISITE=0
+    - php: "7.0"
+      env: WP_VERSION=4.8 WP_MULTISITE=0
+    - php: "7.1"
+      env: WP_VERSION=4.8 WP_MULTISITE=0
+    - php: "5.4"
+      env: WP_VERSION=4.8 WP_MULTISITE=1
     - php: "5.5"
-      env: WP_VERSION=4.4 WP_MULTISITE=0    
-    - php: "5.2"
```
david-binda: Totally get the idea behind dropping PHP 5.2 and am all-in for doing so. Just, could we adjust the comment on line 14 to match this change?

.travis.yml
```diff
@@ -14,34 +18,46 @@ env:
...
-      env: WP_VERSION=master WP_MULTISITE=0
-    - php: "5.2"
-      env: WP_VERSION=4.5 WP_MULTISITE=0
-    - php: "5.2"
-      env: WP_VERSION=4.4 WP_MULTISITE=0
-    - php: "5.3"
```
david-binda: I can also understand the drop of the v 5.3, just again, could we adjust the comment around those test? I might be good to revisit the comment on lines 13-16 and adjust those accordingly.

@sboisvert what are your thoughts on deprecating 5.3 support?

.travis.yml
```diff
@@ -60,7 +76,15 @@ before_script:
     - sed -i "s/yourusernamehere/travis/" wp-tests-config.php
     - sed -i "s/yourpasswordhere//" wp-tests-config.php
     - cd "/tmp/wordpress/src/wp-content/plugins/$PLUGIN_SLUG"
+    - |
+        if [[ ${TRAVIS_PHP_VERSION:0:2} == "7." ]]; then
+          composer global require "phpunit/phpunit=5.7.*"
+        elif [[ ${TRAVIS_PHP_VERSION:0:3} != "5.2" ]]; then
```
david-binda: Is this really needed? Under what circumstances the travis runs in PHP 5.2 environment now, when it was deprecated? I might be missing something, of course :)

raduconst: I added this part because PHPUnit 5 is working starting with PHP 5.6 and for PHP 5.4 and 5.5  the PHPUnit 4 is needed (ref: https://phpunit.de/). The actual error message that appears when not doing this change is:
Package phpunit/phpunit at version 5.7.* has a PHP requirement incompatible with your PHP version (5.4.45)

david-binda: Thanks for the additional details. I perhaps should had been more clear.

The conditional says that the

```
composer global require "phpunit/phpunit=4.8.*"
```

line should be triggered always, when the code runs outside PHP 7.x unless it's PHP 5.2. But, according to the configuration of used PHP versions, there should be no case when the code runs under PHP 5.2. Thus, the conditional could be simple `else`, imho.

.travis.yml
sboisvert: Personally I think we should drop support for < PHP 7. That being said, I don't think that's super realistic. I do think that if we think about who our target audience for these plugins, my thoughts are that they not the kind of people who don't have any control over their PHP version. I think we do need to flag it in our readme and change our docs. And I do think that we shouldn't support version older then x time and be consistent about it.
I personally think that means us testing and supporting 1 year of WordPress and whatever PHP versions are currently supported by the PHP project. We won't say no to patches that fix things for older versions but won't actively test and on them and won't officially support them. http://php.net/supported-versions.php
In the meantime I think we can safely drop 5.2 and 5.3 and strongly consider only supporting 5.6 as the earliest.

david-binda: > We won't say no to patches that fix things for older versions but won't actively test and on them and won't officially support them

In case that's the case, and in case we are going to accept patches for older versions, we should keep those unit tests on those versions, but mark them as Allowed to fail: https://docs.travis-ci.com/user/customizing-the-build/#Rows-that-are-Allowed-to-Fail

For the record, I'm fine with such approach.

sboisvert: >In case that's the case, and in case we are going to accept patches for older versions, we should keep those unit tests on those versions, but mark them as Allowed to fail: https://docs.travis-ci.com/user/customizing-the-build/#Rows-that-are-Allowed-to-Fail

Good idea, I think that's the best of both worlds.

.travis.yml
raduconst: You are right about the else part. When the PHP version is other than 7.x, the phpunit version that can be used is 4.8.*. I'll fo this change in code.

.travis.yml
raduconst: I'm sorry but I don't understand what's the comment I should edit. Is it the one here? https://github.com/Automattic/Edit-Flow/blob/master/.travis.yml#L16

david-binda: I should had provided link right away :) I have been using the line numbers in your brach/fork: https://github.com/raduconst/Edit-Flow/blob/69455a92b45a22727e61f30e2da7ed3be3fa1bd6/.travis.yml#L14

> on minimum PHP 5.2

which is no longer true. We should make sure the comments in the file are accurate. Comments and code should always be consistent and updated along each other, unless future reader can't tell whether the comment is wrong, when clearly saying the opposite of what the code is doing, or whether that's a bug.

With all the changes in the test matrix, there might be other comments which need to be updated, even if only slightly.

.travis.yml
```diff
@@ -62,6 +62,23 @@ matrix:
       env: WP_VERSION=master WP_MULTISITE=0
   allow_failures:
     - php: "nightly"
+  allow_failures:
```
david-binda: Could we, perhaps, merge the two "allow_failures" together? Just another block starts two lines above. Is there a reason for not merging those?

.travis.yml
raduconst: You are right. I don't know how could I miss that. Yes, this is very important to be changed and I think that I can do that based on what we discussed here.

.travis.yml
raduconst: Of course. I was that php nightly was on allowing failure but I forgot to combine it with the new tests.

david-binda:
Thanks for updating the comments and fixing the duplicate allow_failure definition @raduconst ! I can now see the "Allowed Failures" table in https://travis-ci.org/Automattic/Edit-Flow/builds/334642624?utm_source=github_status&utm_medium=notification

However, that only contains the nighly build of PHP. I can't see the PHP 5.2 nor PHP 5.3 in the list of builds on the above link. Feels like the PHP 5.2 and 5.3 should be added to the matrix in terms of include, as what is listed under the allow_failures does not get executed - it's just a list of what and under what conditions should be allowed to fail. Thus, imho, those configurations need to be added under the include, and in terms of allowed_failure we might only need to list the PHP versions which we don't care about - 5.2 and 5.3

raduconst:
@david-binda This is a problem that I could see from the first minute I opened the tests and now I'm looking to what other users said over the internet about this. Yes, I was thinking about the same thing, that I need to add the test in the include section and just mention the versions I want to be allowed to fail. I'm still checking this out and come back with some updates. I'm on it 👍

raduconst:
I think that I finally made it. The tests are running now on 5.2 and 5.3 too but they are on the _Allowed to Fail_ section. The comments are updated too, according to the test configuration. @david-binda Could you have a look again, please?

PR #424, opened 2018-01-23
Prevent user from removing "Draft" post status

### LINKED ISSUE #194: Auto-draft broken when 'draft' doesn't exist as a post status
When a user removes the 'draft' post status, any new post they create doesn't get auto-saved properly. For instance, if they open a post, type some content, and close the window, it should still be accessible in the posts list.

Report: http://wordpress.org/support/topic/auto-draft-to-custom-status-default?replies=2#post-4057764

justnorris:
Fixes #194 

Drafts are a core feature and should be kept intact. This patch will prevent the users to alter both "slug" and "name" of the "Draft" post status, but will allow to edit Draft description and post status order.


modules/custom-status/custom-status.php
```diff
@@ -492,6 +492,16 @@ function update_custom_status( $status_id, $args = array() ) {
...
 		$this->custom_statuses_cache = array();
 
+		// Prevent user from changing draft name or slug
+		if ( 'draft' === $old_status->slug
+		     && (
+			     ( isset( $args['name'] ) && $args['name'] !== $old_status->name )
+			     ||
+			     ( isset( $args['slug'] ) && $args['slug'] !== $old_status->slug )
+		     ) ) {
+			return new WP_Error( 'invalid', __( 'Changing the name and slug of "Draft" is not allowed', 'edit-flow' ) );
```
sboisvert: As a user I think I might end up wondering why this isn't allowed. Perhaps we should put a link to some documentation or at least to the Git Hub issue that talks about it?

modules/custom-status/custom-status.php
```diff
@@ -1830,11 +1840,14 @@ function column_name( $item ) {
...
 		$actions['edit'] = "<a href='$item_edit_link'>" . __( 'Edit', 'edit-flow' ) . "</a>";
 		$actions['inline hide-if-no-js'] = '<a href="#" class="editinline">' . __( 'Quick&nbsp;Edit' ) . '</a>';
-		if ( $item->slug != $this->default_status )
-			$actions['make_default'] = sprintf( '<a href="%1$s">' . __( 'Make&nbsp;Default', 'edit-flow' ) . '</a>', $edit_flow->custom_status->get_link( array( 'action' => 'make-default', 'term-id' => $item->term_id ) ) );
-
-		if ( $item->slug != $this->default_status )
-			$actions['delete delete-status'] = sprintf( '<a href="%1$s">' . __( 'Delete', 'edit-flow' ) . '</a>', $edit_flow->custom_status->get_link( array( 'action' => 'delete-status', 'term-id' => $item->term_id ) ) );
+		$actions['make_default'] = sprintf( '<a href="%1$s">' . __( 'Make&nbsp;Default', 'edit-flow' ) . '</a>', $edit_flow->custom_status->get_link( array( 'action' => 'make-default', 'term-id' => $item->term_id ) ) );
+		
+		// Disable delete/make-default actions for draft status
```
sboisvert: There seems to be an if inside an if but we only do something if they are both true, could we combine them?
I'm also not 100% clear on what this is doing. How are we preventing the make default action?
Why are we preventing draft from being a default status?

modules/custom-status/custom-status.php
justnorris: Unless you "hack the system" you won't see that error. I've disabled input fields for drafts so that you can't actually change the slug and draft title.

modules/custom-status/custom-status.php
justnorris: Bad PHP comment. Initially I disabled making default and came to a conclusion that it shouldn't be prevented, forgot to edit the comment.

sboisvert: The comment is better but we still have nested ifs, not a big deal but it's just strange to read :)

modules/custom-status/custom-status.php
sboisvert: >Unless you "hack the system" you won't see that error. I've disabled input fields for drafts so that you can't actually change the slug and draft title.
Ah, I think I was confused, I thought this would impact current folks who changed it, but it's only when trying to change it.

modules/custom-status/custom-status.php
justnorris: Whoa. After moving all the code around I didn't notice an unnecessary nested if statement. Fixed.

PR #425, opened 2018-01-23
Added a check for the tests dir

raduconst:
When the tests folder is installed locally, the folder might be different than the one used on travis. That's why I added a check. If the path to the tests is found in the environment variables, that one is used. If not, a default path is used. I made this change because the tests were installed on a different folder on my computer and I didn't want to change the $_tests_dir (not even for a test) because I might push it to the plugin (by mistake) and I thought about this workaround.

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -1,6 +1,10 @@
 <?php
 
-$_tests_dir = '/tmp/wordpress/tests/phpunit';
+$_tests_dir = getenv( 'WP_TESTS_DIR' );
+
+if ( !$_tests_dir ) {
+	$_tests_dir = '/tmp/wordpress/tests/phpunit';
+}
 
 require_once $_tests_dir . '/includes/functions.php';
 


sboisvert:
@raduconst Looks good. A few questions:
Why were the tests installed on a different folder on your computer? Is this how it'll be for everyone?
Is WP_TEST_DIR standard? Should we link to where people can find more information about this?

raduconst:
@sboisvert I couldn't find a documentation or something like that. However, I checked some other GitHub repositories and articles over the internet and I could see that the environment variable is named this way and the tests folder is created on this path. Also, this is what I found here: https://github.com/WordPress/phpunit-test-reporter/blob/master/tests/bootstrap.php on a repo from WordPress.

raduconst:
I played some more with tests and I found that this check was available a while back but it was removed in this commit: https://github.com/Automattic/Edit-Flow/commit/247dffb9aa5fd323f52223cad4307964472e1537#diff-10eafd8bf5fbdde867b86fe38d30b9bb
However, I couldn't understand exactly why this change was made.

PR #426, opened 2018-01-24
Fixed the status update dropdown from single post

raduconst:
The dropdown from the Publish block on the right side of the page was displaying the options in just one language. Now everything is translated when using a different language, other than English.

### DIFF
--- a/modules/custom-status/custom-status.php
+++ b/modules/custom-status/custom-status.php
@@ -242,7 +242,7 @@ class EF_Custom_Status extends EF_Module {
 			// of manage posts if there is a post with the status
 			foreach ( $custom_statuses as $status ) {
 				register_post_status( $status->slug, array(
-					'label'       => $status->name
+					'label'       => esc_html__( $status->name, 'edit-flow' )
 					, 'protected'   => true
 					, '_builtin'    => false
 					, 'label_count' => _n_noop( "{$status->name} <span class='count'>(%s)</span>", "{$status->name} <span class='count'>(%s)</span>" )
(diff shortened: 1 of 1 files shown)


modules/custom-status/custom-status.php
```diff
@@ -414,9 +414,9 @@ function post_admin_header() {
 			// Load the custom statuses
 			foreach( $custom_statuses as $status ) {
 				$all_statuses[] = array(
-					'name' => esc_js( $status->name ),
-					'slug' => esc_js( $status->slug ),
-					'description' => esc_js( $status->description ),
+					'name' => esc_html__( $status->name, 'edit-flow' ),
+					'slug' => $status->slug,
```
david-binda: Could you, please, elaborate on why removing the `esc_js` function is safe? I mean, the value used to be escaped by `esc_js` which is now being removed form within a commit claiming that is fixing localisation issues.

This feels like unrelated change, so even if it was safe, it ideally should be a standalone commit.

modules/custom-status/custom-status.php
```diff
@@ -414,9 +414,9 @@ function post_admin_header() {
 			// Load the custom statuses
 			foreach( $custom_statuses as $status ) {
 				$all_statuses[] = array(
-					'name' => esc_js( $status->name ),
-					'slug' => esc_js( $status->slug ),
-					'description' => esc_js( $status->description ),
+					'name' => esc_html__( $status->name, 'edit-flow' ),
```
david-binda: Could you, please, provide more details on why the `esc_js` is being swapped for `esc_html__`. Those two functions are meant for escaping in different context. As the escaping functions are being changed in terms of a changeset which claims to be fixing localisation issues, it might be helpful to have more data on why it's safe to do so.

PR #431, opened 2018-01-25
Fixes the dash between the post title and status label

### LINKED ISSUE #395: No Status Labels Appear Next to Post Title in List
After I enable the custom status feature, all the status labels disappear from the post bowser. 
screen shot 2017-05-30 at 3 47 59 pm

Edit flow is the only plugin I have active. As soon as I disable the plugin, the status labels reappear.

Any help would be much appreciated.

raduconst:
When the custom statuses are used, there's an option to display a column with the statuses on Posts section (on the dashboard). Right now, the dashes between the post title and status label (that are hidden) are still there. The class for post titles was wrong and I replaced the regex with a _substr_ function. This fixes #395.

I checked what happens both when the status column is displayed/hidden and the custom status is not used. Things are working very well.

The problem that this PR is fixing is the one about the dash that exists by default on WordPress between the post/page title and its status:

screenshot at feb 02 14-35-11
When the custom status from Edit Flow is used, a column for statuses is displayed. That's why the status on the end of the post title is no longer needed. However, the dash between is still there:

screenshot at feb 02 14-28-30

According to the comment for the _ef_remove_post_title_trailing_dashes_ function, this code:
`replace(/(.*) - (<span class="post-state".*<\/span>)$/g, "$1$2"));`
should remove this dash:

screenshot at feb 02 14-28-30

However, this is not doing any change so I used this code instead:
`substr (0, jQuery(this).html().indexOf('</a>')));`
that's looking for the end of the anchor and keeps everything before it (including the closing tag).

Also, I changed the selector from https://github.com/Automattic/Edit-Flow/pull/431/files#diff-ffa61788e8ad2e66c3f716f610e629c0L189 because the class of the post title is _page-title_, not _post-title_ and this is the one that should be targeted in order to remove the dash.

### DIFF
--- a/modules/custom-status/lib/custom-status.js
+++ b/modules/custom-status/lib/custom-status.js
@@ -186,8 +186,8 @@ jQuery(document).ready(function() {
 	// Remove the " - " in between a post title and the post-state span (separately hidden via CSS).
 	// This will not affect the dash before post-state-format spans.
 	function ef_remove_post_title_trailing_dashes() {
-		jQuery('.post-title.column-title strong').each(function() {
-			jQuery(this).html(jQuery(this).html().replace(/(.*) - (<span class="post-state".*<\/span>)$/g, "$1$2"));
+		jQuery('.page-title.column-title strong').each(function() {
+			jQuery(this).html(jQuery(this).html().substr (0, jQuery(this).html().indexOf('</a>')));
 		});
 	}
 	


david-binda:
Could you, please, elaborate on what

>The class for post titles was wrong

actually means? This is also used in the commit message. The commit message should give reader an idea of what is being done and quite importantly also why it is being done.

It's not totally clear form the code change why the previous implementation was failing, so more context on that would be helpful when reviewing the code and invaluable in case this change breaks something, and eventually someone else would have to revisit this.

Thanks!

raduconst:
@david-binda What I mean by _The class for post titles was wrong_ is that the selector used in the _modules/custom-status/lib/custom-status.js_ line 189 was 
`.post-title.column-title strong`
and I changed it to 
`.page-title.column-title strong`
It seems that the _page-title_ class is used in all the post titles on the Posts page.

Regarding the regex that I replaced, I thought that this _substr_ change makes things simple. Basically, the code should look for closing anchor tag and keep just the anchor.

sboisvert:
>Regarding the regex that I replaced, I thought that this substr change makes things simple. Basically, the code should look for closing anchor tag and keep just the anchor.

Was there a use case where this wasn't working?

>@david-binda What I mean by The class for post titles was wrong is that the selector used in the modules/custom-status/lib/custom-status.js line 189 was
.post-title.column-title strong
and I changed it to
.page-title.column-title strong
It seems that the page-title class is used in all the post titles on the Posts page.

I think it would be good to edit the original PR description to give more context and elaborate on how this is a CSS class change :)

raduconst:
@sboisvert I just updated the PR description. I think that I should've done this from the first time in order to give some more context about what I changed and why I did that.

PR #438, opened 2018-01-30
Added a test for multisite preview page

raduconst:
This change is meant to test the PR #357. The problem is that when you're previewing a page on a multisite using subdirectory (not subdomain), the link is 

http://eflowmslocal.local/test1?page_id=7&preview_id=7&preview=true

instead of:

http://eflowmslocal.local/test1/?page_id=7&preview_id=7&preview=true

(it's missing the / between test1 and ?page_id).

The solution from the PR is the right one and I want to wrap this up into a test.

Basically, what I'm trying to do is:
1. check if the current installation is multisite;
2. create a new blog; 
3. move to it;
4. create a new page;
5. check its preview link. 

The problem is that the preview link seems to have a broken structure once I create the new blog. This is how the link looks:

http://httpexample.org/sample/?page_id=3&preview=true

The slash appears even though the preview link in the browser is still broken.

### DIFF
--- a/tests/test-edit-flow-custom-status.php
+++ b/tests/test-edit-flow-custom-status.php
@@ -335,7 +335,8 @@ class WP_Test_Edit_Flow_Custom_Status extends WP_UnitTestCase {
 
 	public function test_home_url_slash_available_on_multisite_page_preview() {
 		if ( is_multisite() ) {
-			$blog_id = wpmu_create_blog( site_url(), 'sample', 'test', self::$admin_user_id );
+			global $pagenow;
+			$blog_id = self::factory()->blog->create( array( 'user_id' => self::$admin_user_id ) );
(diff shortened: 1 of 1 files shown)


raduconst:
I managed to test the call for the _home_url_ function from _fix_post_row_actions_.

The second call of the _home_url_ function from _get_preview_link_ is made through a filter and I couldn't reach that point. 
I could take advantage of the fact that the function is public but I don't know what are the parameters ($actions and $post) that I should use. I also thought about using a filter inside the test, like here, for example, https://core.trac.wordpress.org/browser/trunk/tests/phpunit/tests/multisite/network.php#L126 
but I had no luck with that.

raduconst:
Update: once the PR #442 will be merged, this PR will be enough for testing the slash issue on the preview link.

PR #439, opened 2018-01-30
Add local PHPUnit setup

david-binda:
This PR adds standard PHPUnit setup files, as they are generated by the `wp scaffold plugin-tests` command of the [WP CLI project](https://developer.wordpress.org/cli/commands/scaffold/plugin-tests/).

Primarily, this adds the `bin/install-wp-tests.sh` file, which installs the WordPress and test suite to common location. The location is then used in the test/bootstrap.php file.

Further, a multisite.xml config is added to the location which is being mentioned by WordPress PHPUnit test suite for running in MS mode:

```
Running as single site... To run multisite, use -c tests/phpunit/multisite.xml
```

Which `multisite.xml` in that location, following command works like a charm for the phpunit command run from within the plugin:

```
phpunit -c tests/phpunit/multisite.xml
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,18 +89,8 @@ matrix:
     - php: "5.3"
 
 before_script:
-    - export PLUGIN_SLUG=$(basename $(pwd))
-    - git clone --depth=1 --branch $WP_VERSION git://develop.git.wordpress.org/ /tmp/wordpress
-    - cd ..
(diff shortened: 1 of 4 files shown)


david-binda:
I have fixed the Travis tests. @raduconst , feel free to review the PR now, once you get a change. Thanks!

raduconst:
@david-binda I checked this PR (it took me a while to understand what every change is supposed to do + test on my local environment) and everything looks fine. I think that it's ready for merge :-)

david-binda:
Thanks for the thorough review @raduconst !

Issue #440, opened 2018-01-30
Add PHPCS standard to the plugin


david-binda:
As we want to be good citizens, we should fully embrace the WordPress coding standards in terms of this plugin. And, as the plugin is being used on the WordPress.com VIP platforms, we also should be stricter than the standard WordPress coding standards, and fully embrace the VIP and WordPressVIPMinimum standards.

examining the way `wp scaffold plugin my-plugin` ([#](https://developer.wordpress.org/cli/commands/scaffold/plugin-tests/)) generated the PHPCS standards and incorporates it into Travis configuration should give us some starting point for making this true.

Some other related reading:

* https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards
* https://github.com/Automattic/VIP-Coding-Standards

cojennin:
Added with #559

PR #441, opened 2018-01-31
Scope module assets

### LINKED ISSUE #351: Scope loading of module styles and scripts
Sometimes [we check](https://github.com/Automattic/Edit-Flow/blob/master/modules/calendar/calendar.php#L180) if we should load CSS and JS on a given page, [sometimes we don't](https://github.com/Automattic/Edit-Flow/blob/master/modules/calendar/calendar.php#L205).

A patch for this issue would make it so a module's CSS and JS is loaded only when necessary instead of on every page in the admin.

This is consolidating tickets #239 and #242. Exists because of #263. Might also help scope (but not solve) cases like #323.

justnorris:
**Overview**
At the moment, almost all assets are loaded on every admin page (for example - `calendar.js` is loaded in "Settings -> General”). Overall there were a lot of inconsistencies in asset enqueuing. The goal of this PR is to provide a reliable and predictable way of enqueuing module assets.

**Background**
Every Edit Flow module extends the `EF_Module`  class. However, some modules have dedicated views and some modules (like Dashboard widgets module) don’t. That’s where the inconsistencies begin. 

Most EF Modules enqueue assets. Some check whether they should enqueue assets, some perform the same checks in a different way, and some don’t check at all. Most of the modules use method `EF_Module::is_whitelisted_functional_view` which simply always returns true, with a `//@todo` attached to it :) - the whole asset enqueuing process needed a refactor. 

**Introducing** `EF_Module_With_View`
The EF modules needed a few methods to make it easy to check whether the any given module is visible at the moment (and therefore, whether they should enqueue assets). In most cases this is either in the edit, list and settings views. 

At first I dumped all the necessary methods on `EF_Module` class because all of the modules already are extending it. But `EF_Module` is also instantiated directly **and** it’s also extended by a `EF_Dashboard` class that doesn’t utilize the any of the “new” methods, so it didn’t make sense to keep them there. 

`EF_Module_With_View extends EF_Module` and is meant to be extended further by modules that have views, that way they have access to necessary methods, like `is_current_module_settings_view()` and `is_active_view()`

**Adding PHP interfaces**
There was no streamlined way of dealing with assets. Even enqueuing method names varied from module to module. That’s why I decided to 2 simple interfaces: `EF_Script_Interface` and  `EF_Style_Interface` - they will ensure that asset enqueuing methods are consistent and predictable. Also, by reading the class declaration it’s immediately clear whether or not the current module is enqueuing any assets.

**Summary**
* Force predictable and consistent enqueuing methods with PHP Interfaces across EF Modules
* Add a an reusable and easy to use class `EF_Module_With_View` that provides methods to easily determine whether or not assets are needed
* Refactor most modules to extend `EF_Module_With_View` instead of  `EF_Module` and enqueue assets when they’re needed

Fixes #351

### DIFF
--- a/common/php/class-module-with-view.php
+++ b/common/php/class-module-with-view.php
@@ -1,0 +1,181 @@
+<?php
+
+
+class EF_Module_With_View extends EF_Module {
+
+
+	/**
+	 * Whether or not the current page is an Edit Flow settings view (either main or module)
+	 * Determination is based on $pagenow, $_GET['page'], and the module's $settings_slug
+	 * If there's no module name specified, it will return true against all Edit Flow settings views
+	 *
+	 * @since 0.8.3
+	 *
+	 * @param string $slug (Optional) Module name to check against
+	 * @return bool true if is module settings view
+	 */
+	public function is_module_settings_view( $slug = false ) {
+		global $pagenow, $edit_flow;
+
+		// All of the settings views are based on admin.php and a $_GET['page'] parameter
+		if ( $pagenow !== 'admin.php' || ! isset( $_GET['page'] ) )
+			return false;
+
+		$settings_view_slugs = array();
+		// Load all of the modules that have a settings slug/ callback for the settings page
+		foreach ( $edit_flow->modules as $mod_name => $mod_data ) {
+			if ( isset( $mod_data->options->enabled ) && $mod_data->options->enabled == 'on' && $mod_data->configure_page_cb )
+				$settings_view_slugs[] = $mod_data->settings_slug;
+		}
+
+		// The current page better be in the array of registered settings view slugs
+		if ( empty( $settings_view_slugs ) || ! in_array( $_GET['page'], $settings_view_slugs ) ) {
+			return false;
+		}
+
+		if ( $slug && $edit_flow->modules->{$slug}->settings_slug !== $_GET['page'] ) {
+			return false;
+		}
+
+		return true;
+	}
+
+	/**
+	 * Check whether if we're at module settings view
+	 * for the current module based on `is_module_settings_view` method
+	 *
+	 * @return bool
+	 */
+	public function is_current_module_settings_view() {
+		return $this->is_module_settings_view( $this->module->name );
+	}
+
+
+	/**
+	 * Check for admin page and whether the current post type is supported
+	 * @param array $allowed_pages
+	 *
+	 * @return bool
+	 */
+	function is_active_view( $allowed_pages = array( 'edit.php', 'post.php', 'post-new.php' ) ) {
+		return ( $this->is_admin_page( $allowed_pages ) && $this->is_supported_post_type() );
+	}
+
+
+	/**
+	 * Check whether the current post type is supported for $this module
+	 *
+	 * @return bool
+	 */
+	public function is_supported_post_type() {
+		$post_type = $this->get_current_post_type();
+
+		return (
+			$post_type
+			&&
+			in_array( $post_type, $this->get_post_types_for_module( $this->module ), true )
+		);
+	}
+
+	/**
+	 * Checks for the current post type
+	 *
+	 * @since 0.7
+	 * @return string|null $post_type The post type we've found, or null if no post type
+	 */
+	function get_current_post_type() {
+		global $post, $typenow, $pagenow, $current_screen;
+		//get_post() needs a variable
+		$post_id = isset( $_REQUEST['post'] ) ? (int) $_REQUEST['post'] : false;
+
+		if ( $post && $post->post_type ) {
+			$post_type = $post->post_type;
+		} elseif ( $typenow ) {
+			$post_type = $typenow;
+		} elseif ( $current_screen && ! empty( $current_screen->post_type ) ) {
+			$post_type = $current_screen->post_type;
+		} elseif ( isset( $_REQUEST['post_type'] ) ) {
+			$post_type = sanitize_key( $_REQUEST['post_type'] );
+		} elseif ( 'post.php' == $pagenow
+		           && $post_id
+		           && ! empty( get_post( $post_id )->post_type ) ) {
+			$post_type = get_post( $post_id )->post_type;
+		} elseif ( in_array( $pagenow, array('edit.php', 'post-new.php' ) ) && empty( $_REQUEST['post_type'] ) ) {
+			$post_type = 'post';
+		} else {
+			$post_type = null;
+		}
+
+		return $post_type;
+	}
+
+	/**
+	 * Check whether currently viewing the desired admin page
+	 *
+	 * @param array $allowed_pages
+	 *
+	 * @return bool
+	 */
+	public function is_admin_page( $allowed_pages = array( 'edit.php', 'post.php', 'post-new.php' ) ) {
+		global $pagenow;
+
+		return ( $pagenow && in_array( $pagenow, $allowed_pages, true ) );
+	}
+
+
+	/**
+	 * Shorthand for `is_active_view` to check for list type views ( list of posts pages, custom post types )
+	 *
+	 * @see is_active_view
+	 * @return bool
+	 */
+	public function is_active_list_view() {
+		return $this->is_active_view( array( 'edit.php' ) );
+	}
+
+	/**
+	 * Shorthand for `is_active_view` to check for editor mode
+	 *
+	 * @see is_active_view
+	 * @return bool
+	 */
+	public function is_active_editor_view() {
+		return $this->is_active_view( array( 'post.php', 'posts-new.php' ) );
+	}
+
+
+	/**
+	 * This method was supposed to check whether or not the current page is a user-facing Edit Flow View
+	 * But it was never implemented
+	 *
+	 * It is now deprecated in favor of `$this->is_active_view`
+	 * @see        EF_Module::is_active_view()
+	 * @since      0.7
+	 * @deprecated 0.8.3
+	 *
+	 * @param string $module_name (Optional) Module name to check against
+	 */
+	public function is_whitelisted_functional_view( $module_name = null ) {
+		_deprecated_function( __FUNCTION__, '0.8.3', 'is_active_view' );
+
+		return true;
+	}
+
+	/**
+	 * Whether or not the current page is an Edit Flow settings view (either main or module)
+	 * Determination is based on $pagenow, $_GET['page'], and the module's $settings_slug
+	 * If there's no module name specified, it will return true against all Edit Flow settings views
+	 *
+	 * @since 0.7
+	 * @deprecated 0.8.3
+	 *
+	 * @param string $module_name (Optional) Module name to check against
+	 * @return bool $is_settings_view Return true if it is
+	 */
+	public function is_whitelisted_settings_view( $module_name = null ) {
+		_deprecated_function( 'is_whitelisted_settings_view', '0.8.3', 'is_module_settings_view' );
+
+		return $this->is_module_settings_view( $module_name );
+	}
+
+}
--- a/modules/calendar/calendar.php
+++ b/modules/calendar/calendar.php
@@ -782,10 +782,14 @@ class EF_Calendar extends EF_Module {
 						$this->hidden = 0;
 						if ( !empty( $week_posts[$week_single_date] ) ) {
 
-							$week_posts[$week_single_date] = apply_filters( 'ef_calendar_posts_for_week', $week_posts[$week_single_date] );
+							$week_posts[$week_single_date] = apply_filters( 'ef_calendar_posts_for_week', $week_posts[$week_single_date], $week_single_date );
 
-							foreach ( $week_posts[$week_single_date] as $num => $post ){ 
-								echo $this->generate_post_li_html( $post, $week_single_date, $num ); 
+							foreach ( $week_posts[$week_single_date] as $num => $post ) {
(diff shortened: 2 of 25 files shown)


sboisvert:
Wow! This is a great PR and great code!
It's also a lot of changes! That's not necessarily a bad thing but it does mean a lot of testing and a lot of potential stuff that could break.

A few questions that came to mind while reading the code and the PR:
What could this potentially break? 
If people are extending EF_Module, does the presence of EF_Module_With_View change things?
What about the functions we're removing from EF_Module?
Are these legitimate concerns or just some overly broad concerns that probably won't happen in the "real world"?
How are the interfaces making the code easier to read / more sustainable? What other solutions could there be that we didn't pick and why?
Should we add more comments about the interfaces and their roles? I don't see them used very often in WP code and I suspect there might be many WP coders who do not have experience with them.

justnorris:
> What could this potentially break?

> What about the functions we're removing from EF_Module?

> Are these legitimate concerns or just some overly broad concerns that probably won't happen in the "real world"?

There are 3 scenarios that I can imagine where things might break.

1. The first is that someone has dequeued styles or scripts of a module. I've kept all the original script handles, but for some, I had to change the when they're enqueued for consistency. For example, EF_Editorial_Metadata was enqueued on `add_admin_scripts` hook instead of `admin_enqueue_scripts` where it should be. I don't think that that in reality this is going to break something, and even if it does - it's just a hook change.

Where people might have used hooks would be to turn off `EF_Editorial_Metadata` styles/scripts in places where they don't want the styles/scripts - which is what this PR does already, which is why I changed the hooks a tiny bit.

2. If someone has extended the `EF_Module` class **and** used the `is_whitelisted_settings_view` method - that method is no longer there. This is most worrisome, but I don't think it's a real scenario either. If it is - the fix is easy - just extend `EF_Module_With_View` and use `is_current_module()` method instead.

To remedy this, I can add a fallback `is_whitelisted_settings_view` method on the `EF_Module` class that returns true and shows a deprecation warning. It would break the functionality, but counter a hypothetical fatal error.

3. If someone has been relying on `EF_Module::get_current_post_type()` method. Same as (2) - it's moved to `EF_Module_With_View`. I don't think it's actually used anywhere to extend the class, but I can add the fallback to counter any hypothetical fatal error.


> If people are extending EF_Module, does the presence of EF_Module_With_View change things?
No, unless (2) and (3). 

> How are the interfaces making the code easier to read / more sustainable? What other solutions could there be that we didn't pick and why?

Interfaces are entirely optional, but provide a way to predict what the code below does without having to go through methods. An alternative would be to simply remove them - the newer method names will remain the same anyway.  Interfaces have been around since PHP 5.0 and these days they are used in the wider PHP Community, quite heavily in the Laravel community at least. Even if a developer isn't used to interfaces, I think it's quite straight forward to guess what they do.

I can remove them to avoid that confusion. I used them there when refactoring to keep myself from inventing a new enqueuing assets method for each class or having to go back each time to find what was the method name I am supposed to use. Not using interfaces is what caused code inconsistencies across modules in the first place, and imho - we should be using features that the language provides.


> Should we add more comments about the interfaces and their roles? I don't see them used very often in WP code and I suspect there might be many WP coders who do not have experience with them.
Do you have any suggestions on what that comment might look like? I'm used to interfaces and I thought them straight forward without having to write a larger comment. 



------

All in all - there are a lot of changes, but no that much can break. This PR is enqueuing assets when they're needed and removing 3 likely unused methods and changing the hook to the correct hook for enqueuing admin scripts/styles. The 3 methods aren't "really gone" they're just renamed, so it's easy to fix if anyone has actually relied on those methods for a plugin extension.

sboisvert:
@philipjohn Could you give this a look over or assign it to someone if possible?

common/php/class-module-with-view.php
```diff
@@ -0,0 +1,181 @@
...
+	 * @since 0.8.3
+	 *
+	 * @param string $slug (Optional) Module name to check against
+	 * @return bool true if is module settings view
+	 */
+	public function is_module_settings_view( $slug = false ) {
+		global $pagenow, $edit_flow;
+
+		// All of the settings views are based on admin.php and a $_GET['page'] parameter
+		if ( $pagenow !== 'admin.php' || ! isset( $_GET['page'] ) )
```
emrikol: Braces should be added to match [WordPress PHP Coding Standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#brace-style)

common/php/class-module-with-view.php
```diff
@@ -0,0 +1,181 @@
...
+			return false;
+
+		$settings_view_slugs = array();
+		// Load all of the modules that have a settings slug/ callback for the settings page
+		foreach ( $edit_flow->modules as $mod_name => $mod_data ) {
+			if ( isset( $mod_data->options->enabled ) && $mod_data->options->enabled == 'on' && $mod_data->configure_page_cb )
```
emrikol: What about switching to a yoda condition and a strict check?

`'on' === $mod_data->options->enabled`

common/php/class-module-with-view.php
```diff
@@ -0,0 +1,181 @@
...
+				$settings_view_slugs[] = $mod_data->settings_slug;
+		}
+
+		// The current page better be in the array of registered settings view slugs
+		if ( empty( $settings_view_slugs ) || ! in_array( $_GET['page'], $settings_view_slugs ) ) {
```
emrikol: `in_array()` should have the third argument as `true` for a strict check, unless we specifically don't want one here.

common/php/class-module-with-view.php
emrikol: Braces should be added to match [WordPress PHP Coding Standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#brace-style)

common/php/class-module-with-view.php
```diff
@@ -0,0 +1,181 @@
...
+	/**
+	 * Checks for the current post type
+	 *
+	 * @since 0.7
+	 * @return string|null $post_type The post type we've found, or null if no post type
+	 */
+	function get_current_post_type() {
+		global $post, $typenow, $pagenow, $current_screen;
+		//get_post() needs a variable
+		$post_id = isset( $_REQUEST['post'] ) ? (int) $_REQUEST['post'] : false;
```
emrikol: Do we need `$_REQUEST` here or can we be more specific with `$_GET` or `$_POST`?

common/php/class-module-with-view.php
```diff
@@ -0,0 +1,181 @@
...
+
+		if ( $post && $post->post_type ) {
+			$post_type = $post->post_type;
+		} elseif ( $typenow ) {
+			$post_type = $typenow;
+		} elseif ( $current_screen && ! empty( $current_screen->post_type ) ) {
+			$post_type = $current_screen->post_type;
+		} elseif ( isset( $_REQUEST['post_type'] ) ) {
```
emrikol: Do we need `$_REQUEST` here or can we be more specific with `$_GET` or `$_POST`?

common/php/class-module-with-view.php
```diff
@@ -0,0 +1,181 @@
...
+			$post_type = sanitize_key( $_REQUEST['post_type'] );
```
emrikol: Do we need `$_REQUEST` here or can we be more specific with `$_GET` or `$_POST`?

common/php/class-module-with-view.php
```diff
@@ -0,0 +1,181 @@
...
+		} elseif ( 'post.php' == $pagenow
+		           && $post_id
+		           && ! empty( get_post( $post_id )->post_type ) ) {
+			$post_type = get_post( $post_id )->post_type;
+		} elseif ( in_array( $pagenow, array('edit.php', 'post-new.php' ) ) && empty( $_REQUEST['post_type'] ) ) {
```
emrikol: Do we need `$_REQUEST` here or can we be more specific with `$_GET` or `$_POST`?

emrikol: `in_array()` should have the third argument as `true` for a strict check, unless we specifically don't want one here.

modules/calendar/calendar.php
```diff
@@ -1836,7 +1841,13 @@ public function fix_post_date_on_update_part_two( $post_ID, $post_after, $post_b
 		$wpdb->update( $wpdb->posts, array( 'post_date' => $post_date ), array( 'ID' => $post_ID ) );
 		clean_post_cache( $post_ID );
 	}
-	
+
+
+	public function is_calendar_view() {
+		global $pagenow;
+
+		return ( 'index.php' === $pagenow && isset( $_GET['page'] ) && $_GET['page'] === 'calendar' );
```
emrikol: Can this be switched to a yoda conditional?

`'calendar' === $_GET['page']`

modules/editorial-comments/editorial-comments.php
```diff
@@ -80,27 +84,31 @@ function upgrade( $previous_version ) {
...
+		}
 
 		wp_enqueue_script( 'edit_flow-post_comment', $this->module_url . 'lib/editorial-comments.js', array( 'jquery','post' ), EDIT_FLOW_VERSION, true );
-		wp_enqueue_style( 'edit-flow-editorial-comments-css', $this->module_url . 'lib/editorial-comments.css', false, EDIT_FLOW_VERSION, 'all' );
 
 		$thread_comments = (int) get_option('thread_comments');
 		?>
 		<script type="text/javascript">
-			var ef_thread_comments = <?php echo ($thread_comments) ? $thread_comments : 0; ?>;
+            var ef_thread_comments = <?php echo ($thread_comments) ? $thread_comments : 0; ?>;
```
emrikol: We should escape this PHP output with `wp_json_encode()`

modules/story-budget/story-budget.php
```diff
@@ -133,36 +133,43 @@ function upgrade( $previous_version ) {
...
-		
-		$num_columns = $this->get_num_columns();
-		echo '<script type="text/javascript"> var ef_story_budget_number_of_columns="' . esc_js( $this->num_columns ) . '";</script>';
-		
-		$this->enqueue_datepicker_resources();
-		wp_enqueue_script( 'edit_flow-story_budget', $this->module_url . 'lib/story-budget.js', array( 'edit_flow-date_picker' ), EDIT_FLOW_VERSION, true );
+
+		if ( $this->is_story_board_view() ) {
+			$num_columns = $this->get_num_columns();
+			echo '<script type="text/javascript"> var ef_story_budget_number_of_columns="' . esc_js( $this->num_columns ) . '";</script>';
```
emrikol: We should use `wp_json_encode()` (watch for double quotes) instead of `esc_js()`

justnorris:
@emrikol My goal with this PR was to fix #351 with as little change as possible. And even so - I had to change a lot of things. I agree with all of the code improvements you mentioned, but perhaps that should be a separate PR? 

I think a good first step towards that would be filling #440. 

What are your thoughts? Do you want me to go ahead and fix the code style issues, or leave them be for #440?

Sidenote: The issues you found were already there, I just moved the code - that's the reason I'm hesitant to fix them in the PR. If I fix code style issues with this PR it might be a lot more difficult to backtrack for someone looking for what happened and how the changes were made.

common/php/class-module-with-view.php
justnorris: I tried to keep this PR short so I just moved the code around without updating the PHCPS standards support to keep it short. I agree that we should work on this, but let's leave that for a different PR.

common/php/class-module-with-view.php
justnorris: Same as https://github.com/Automattic/Edit-Flow/pull/441/files#r240276539 :)

common/php/class-module-with-view.php
justnorris: Same as https://github.com/Automattic/Edit-Flow/pull/441/files#r240276539 :)

common/php/class-module-with-view.php
justnorris: Same as https://github.com/Automattic/Edit-Flow/pull/441/files#r240276539 :)

common/php/class-module-with-view.php
justnorris: Same as https://github.com/Automattic/Edit-Flow/pull/441/files#r240276539 :)

common/php/class-module-with-view.php
justnorris: Same as https://github.com/Automattic/Edit-Flow/pull/441/files#r240276539 :)

common/php/class-module-with-view.php
justnorris: Same as https://github.com/Automattic/Edit-Flow/pull/441/files#r240276539 :)

justnorris: Same as https://github.com/Automattic/Edit-Flow/pull/441/files#r240276539 :)

modules/story-budget/story-budget.php
justnorris: Same as https://github.com/Automattic/Edit-Flow/pull/441/files#r240276539 :)

modules/calendar/calendar.php
justnorris: Same as https://github.com/Automattic/Edit-Flow/pull/441/files#r240276539 :)

justnorris:
Oh I messed this one up good, re-opening in #487 :)

PR #442, opened 2018-02-02
Refactor for the fix_post_row_actions function

raduconst:
There code from lines 1649 to 1664 on custom-status.php file is the same as the one from _get_preview_link_. That's why I replaced it with the call of the _get_preview_link_. 
Another change made in this PR is making this _get_preview_link_ method static. The purpose of this function is to change the actions of a post and it's not related to the class itself. Making it static will let us use it outside this _EF_Custom_Status_ class at any moment.

### DIFF
--- a/modules/custom-status/custom-status.php
+++ b/modules/custom-status/custom-status.php
@@ -1611,7 +1611,6 @@ class EF_Custom_Status extends EF_Module {
 		} else if ( 'post' == $post->post_type ) {
 			$args = array(
 					'p'          => $post->ID,
-					'preview'	 => 'true'
 				);
 		} else {
 			$args = array(
@@ -1620,7 +1619,7 @@ class EF_Custom_Status extends EF_Module {
 				);
 		}
 
-		$args['preview'] = $post->ID;
+		$args['preview_id'] = $post->ID;
 		return add_query_arg( $args, home_url() );
 	}
 
--- a/modules/custom-status/custom-status.php
+++ b/modules/custom-status/custom-status.php
@@ -1602,7 +1602,7 @@ class EF_Custom_Status extends EF_Module {
 	 *
 	 * @since 0.8
 	 */
-	private function get_preview_link( $post ) {
+	static function get_preview_link( $post ) {
 
 		if ( 'page' == $post->post_type ) {
 			$args = array(
@@ -1646,22 +1646,7 @@ class EF_Custom_Status extends EF_Module {
 		if ( empty( $actions['view'] ) )
 			return $actions;
 
-		if ( 'page' == $post->post_type ) {
-			$args = array(
-					'page_id'    => $post->ID,
-				);
-		} else if ( 'post' == $post->post_type ) {
-			$args = array(
-					'p'          => $post->ID,
-				);
-		} else {
-			$args = array(
-					'p'          => $post->ID,
-					'post_type'  => $post->post_type,
-				);
-		}
-		$args['preview'] = 'true';
-		$preview_link = add_query_arg( $args, home_url() );
+		$preview_link = $this->get_preview_link($post);
 
 		$actions['view'] = '<a href="' . esc_url( $preview_link ) . '" title="' . esc_attr( sprintf( __( 'Preview &#8220;%s&#8221;' ), $post->post_title ) ) . '" rel="permalink">' . __( 'Preview' ) . '</a>';
 		return $actions;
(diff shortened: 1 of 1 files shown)


modules/custom-status/custom-status.php
```diff
@@ -1602,7 +1602,7 @@ function fix_get_sample_permalink_html( $return, $post_id, $new_title, $new_slug
 	 *
 	 * @since 0.8
 	 */
-	private function get_preview_link( $post ) {
+	static function get_preview_link( $post ) {
```
david-binda: I know that the visibility declaration of methods in this file is not uniform, but we, imho, should follow the best practices declared in the [WordPress coding standards project](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/805346356689075d9f40b57cbedd05bc813eb5ba/WordPress-Extra/ruleset.xml#L93), specifically the WordPress-Extra ruleset, which sets up best practices beyond the core WordPress coding standards.

And thus, the `public` visibility declaration should be used here. It would also help to highlight the change, as this method becomes public by removing the `private` visibility.

raduconst: Thank you for the review!

I checked the [WordPress Coding Standards](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/1101)  and I can see that the `Methods: Verify that all methods have visibility declared`. Does this mean that I should use

`public function get_preview_link( $post )`

instead of

`function get_preview_link( $post )`

am I right? I just want to make sure that I follow the Coding Standards and write this the right way.

david-binda: Right. `public static function get_preview_link( $post )` to be exact.

raduconst:
@david-binda Done, I made the change. Could you have another look, please? Thanks!

modules/custom-status/custom-status.php
```diff
@@ -1646,22 +1646,7 @@ public function fix_post_row_actions( $actions, $post ) {
...
-			$args = array(
-					'p'          => $post->ID,
-				);
-		} else {
-			$args = array(
-					'p'          => $post->ID,
-					'post_type'  => $post->post_type,
-				);
-		}
-		$args['preview'] = 'true';
```
david-binda: Shall we preserve the `preview` arg for all the links, not only for the case `$post->post_type` is `'post'`? Looks like it might be safer. The `get_preview_link` function currently adds it in the above case, while here it's always used.

raduconst: Yes, I think that it's a good idea. I thought that I already let this arg to `preview` but I think I rushed with the commit and forgot to check this once again. I'll do the change shortly.

raduconst:
@david-binda Can you have another look at this, please?

modules/custom-status/custom-status.php
```diff
@@ -1620,7 +1620,7 @@ private function get_preview_link( $post ) {
 				);
 		}
 
-		$args['preview_id'] = $post->ID;
```
david-binda: Let's keep the `preview_id` as it was/is for all links.

modules/custom-status/custom-status.php
```diff
@@ -1620,7 +1620,7 @@ private function get_preview_link( $post ) {
...
+		$args['preview'] = $post->ID;
```
david-binda: Let's remove the `preview` param from line [1614](https://github.com/Automattic/Edit-Flow/pull/442/files#diff-228291de10e40133681b1ff8414a9a54R1614) if we are adding it here (which imho we should)

raduconst: I made this change right now. I understand why line 1614 can be removed (it was so obvious once I checked the whole method once again).

rinatkhaziev:
@david-binda Do you still have questions regarding this PR or I can mark it for 0.8.3 merge?

david-binda:
@rinatkhaziev @raduconst it still looks like the `preview` query arg is being removed in terms of this PR.

My understanding of the code says it should be preserved. @raduconst any change you could get into restoring the `preview` link in term of the `get_preview_link` method? Eg.: by adding `$args['preview'] = true;` to https://github.com/Automattic/Edit-Flow/pull/442/files#diff-228291de10e40133681b1ff8414a9a54R1622 ?

@rinatkhaziev , feel free to correct me, if my understanding is off.

raduconst:
I'm on my way of checking this out and see what can I do about it. I'll come back with some more info once I look at this once again.

PR #443, opened 2018-02-05
Added checks for `add_screen_options_panel()`

### LINKED ISSUE #354: convert_to_screen(), add_meta_box() was called incorrectly?
I get the following error on every front-end page-load. 4 times, specifically, in groups of two.

I've traced the error back to Edit Flow, but since I don't understand it I don't really know what to do with it. Can I ignore it?

`PHP Notice:  convert_to_screen(), add_meta_box() was called <strong>incorrectly</strong>. Likely direct inclusion of wp-admin/includes/template.php in order to use add_meta_box(). This is very wrong. Hook the add_meta_box() call into the add_meta_boxes action instead. Please see <a href="https://codex.wordpress.org/Debugging_in_WordPress">Debugging in WordPress</a> for more information. (This message was added in version 3.3.) in .../wp-includes/functions.php on line 3897`

shantanu2704:
Fixes #354 

Added `is_admin() conditional` before making a call to `add_screen_options_panel()` in _calendar_ and _custom-budget_ modules since the `screen-options` is being accessed on the front-end.

 I've mentioned two other approaches for this issue on issue page. If you find those are better, let me know and I'll make the changes.

### DIFF
--- a/modules/calendar/calendar.php
+++ b/modules/calendar/calendar.php
@@ -86,8 +86,11 @@ class EF_Calendar extends EF_Module {
 		// Define the create-post capability
 		$this->create_post_cap = apply_filters( 'ef_calendar_create_post_cap', 'edit_posts' );
 		
-		require_once( EDIT_FLOW_ROOT . '/common/php/' . 'screen-options.php' );
-		add_screen_options_panel( self::usermeta_key_prefix . 'screen_options', __( 'Calendar Options', 'edit-flow' ), array( $this, 'generate_screen_options' ), self::screen_id, false, true );
+		if ( is_admin() ) {
+				require_once( EDIT_FLOW_ROOT . '/common/php/' . 'screen-options.php' );
+				add_screen_options_panel( self::usermeta_key_prefix . 'screen_options', __( 'Calendar Options', 'edit-flow' ), array( $this, 'generate_screen_options' ), self::screen_id, false, true );
+		}
+		
 		add_action( 'admin_init', array( $this, 'handle_save_screen_options' ) );
 		
 		add_action( 'admin_init', array( $this, 'register_settings' ) );
--- a/modules/story-budget/story-budget.php
+++ b/modules/story-budget/story-budget.php
@@ -68,9 +68,11 @@ class EF_Story_Budget extends EF_Module {
 		
 		add_action( 'admin_init', array( $this, 'handle_form_date_range_change' ) );
 		
-		include_once( EDIT_FLOW_ROOT . '/common/php/' . 'screen-options.php' );
-		if ( function_exists( 'add_screen_options_panel' ) )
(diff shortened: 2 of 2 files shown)


sboisvert:
Hi @shantanu2704!

Thanks for the patch! 
Can you elaborate on why you chose this solution over the other ones? 
Are there any downsides to this solution?

Thanks!

shantanu2704:
Hey @sboisvert!

The other approaches would need a lot of refactoring in all the modules, not just the calendar and story-budget modules,  to maintain uniformity across the codebase and I didn't want to do those without feedback from you. But I would be willing to do it if I have your approval. I think it would be a good approach in the long run.
This approach is more of a proof of concept. The downside to it is that if someone were to use screen-options in other modules, the use of the `is_admin()` conditional won't be that obvious unless they are familiar with this issue.

sboisvert:
Thanks Shantanu!
I'm not sure I understood what you meant by:
>The downside to it is that if someone were to use screen-options in other modules, the use of the is_admin() conditional won't be that obvious unless they are familiar with this issue.

shantanu2704:
What I mean is if in the future we wanted to add more settings under
screen-options in any other module, this issue could occur again if the
author forgets to check for `is_admin` or is unfamiliar with the current
issue.
But if the actions firing on WP init and dmin_init were separated, it
becomes apparent for anyone using screen-options that it must only be used
on the back-end.

shantanu2704:
Hey @sboisvert!
I'm sorry I wasn't clear about the shortcomings earlier. I hope you understand what I meant.
Another reason that I suggested refactoring the code from `init()` into `init()` and `admin_init()` is:

Currently, the following lines in the `init()` function are being called on the front-end as well as the back-end since WP `init` runs on both. 

https://github.com/Automattic/Edit-Flow/blob/f52a2eb013faf9c75f3b4410285f892b05d82a27/modules/calendar/calendar.php#L88-L96

But this code should run only in the admin area. It doesn't really cause any issue except for the `add_screen_options_panel()` function call.

By refactoring the function, we would avoid unnecessary calls at run-time.

modules/calendar/calendar.php
```diff
@@ -86,8 +86,11 @@ function init() {
 		// Define the create-post capability
 		$this->create_post_cap = apply_filters( 'ef_calendar_create_post_cap', 'edit_posts' );
 		
-		require_once( EDIT_FLOW_ROOT . '/common/php/' . 'screen-options.php' );
-		add_screen_options_panel( self::usermeta_key_prefix . 'screen_options', __( 'Calendar Options', 'edit-flow' ), array( $this, 'generate_screen_options' ), self::screen_id, false, true );
+		if ( is_admin() ) {
```
shantanu2704: The lack of this conditional caused this issue. 
If we were to use  `add_screen_options_panel` in the future in some other module, the current issue would reoccur.

rinatkhaziev:
@shantanu2704 once again, thanks for your work. Since you've not responded in #354 I went ahead and opened a #463 which solves the issue by moving the logic from `init` to `admin_init`

PR #446, opened 2018-03-20
Fix status list, calendar, and story budget to show custom post statuses and show post states in post list.

### LINKED ISSUE #445: Custom post statuses added via filter are not added to status column
If I add a custom post status via `display_post_states`, it is not added to the custom status column, even though it is appended to the post title ( but hidden )

mikeyarce:
Fixes #445

Previously, the `get_post_status_friendly_name` function was retrieving the WordPress built-in Post Statuses and the Custom Statuses set within the plugin interface.  If you happen to register a post status another way, it would not appear in the Status Column of your Post Type.

This PR removes the need for `get_post_status_friendly_name` in the Post List, Calendar, and Story Budget sections and instead just calls the post status label directly.

This also adds a check to see if a Post State does not match the Post Status, and if it does not match, it will show the Post State.  Example:
screen shot 2018-03-20 at 9 32 21 pm

### DIFF
--- a/modules/notifications/notifications.php
+++ b/modules/notifications/notifications.php
@@ -551,9 +551,9 @@ jQuery(document).ready(function($) {
 			/* translators: 1: date, 2: time, 3: timezone */
 			$body .= sprintf( __( 'This action was taken on %1$s at %2$s %3$s', 'edit-flow' ), date_i18n( get_option( 'date_format' ) ), date_i18n( get_option( 'time_format' ) ), get_option( 'timezone_string' ) ) . "\r\n";
 			
-			$old_status_friendly_name = $this->get_post_status_friendly_name( $old_status );
-			$new_status_friendly_name = $this->get_post_status_friendly_name( $new_status );
-			
+			$old_status_friendly_name = get_post_status_object( $old_status )->label;
+			$new_status_friendly_name = get_post_status_object( $new_status )->label;
+						
 			// Email body
 			$body .= "\r\n";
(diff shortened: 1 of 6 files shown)


pdewouters:
@mikeyarce thanks for the quick fix! this seems to work fine.

modules/notifications/notifications.php
```diff
@@ -550,10 +550,14 @@ function notification_status_change( $new_status, $old_status, $post ) {
 			
 			/* translators: 1: date, 2: time, 3: timezone */
 			$body .= sprintf( __( 'This action was taken on %1$s at %2$s %3$s', 'edit-flow' ), date_i18n( get_option( 'date_format' ) ), date_i18n( get_option( 'time_format' ) ), get_option( 'timezone_string' ) ) . "\r\n";
-			
-			$old_status_friendly_name = $this->get_post_status_friendly_name( $old_status );
-			$new_status_friendly_name = $this->get_post_status_friendly_name( $new_status );
-			
+
+			if ( $old_status == 'new' ) {
+				return;
```
anigeluk: Instead of returning here if the old status is 'new' we should still allow a notification to be sent. We have already created a subject to handle this situation in line 514.

If we were to supply an `$old_status_friendly_name` for this instance it would prevent the call to `get_post_status_object` in line 557 failing and causing an error for new posts.

sboisvert:
@anigeluk @mikeyarce LGTM after the change, merging

PR #448, opened 2018-04-03
Prevent undefined index error during subscriptions saving

WPprodigy:
1) Fixes : `PHP Notice:  Undefined index: …/plugins/Edit-Flow/modules/notifications/notifications.php on line 370`.
2) Fixes: `PHP Notice:  Undefined index: _nonce in .../plugins/Edit-Flow/modules/notifications/notifications.php on line 365`

2) Adds further checks to make sure all values are set before accessing them.

3) General cleanup and coding standards for the method.

______

To replicate the error and test this patch:

1) Add some user group subscriptions to a post.
2) Remove all groups from the subscriptions.
3) Note the php notice in the logs.

### DIFF
--- a/modules/notifications/notifications.php
+++ b/modules/notifications/notifications.php
@@ -360,25 +360,31 @@ jQuery(document).ready(function($) {
 	 */
 	function ajax_save_post_subscriptions() {
 		global $edit_flow;
-		
+
 		// Verify nonce.
 		if ( ! isset( $_POST['_nonce'] ) || ! wp_verify_nonce( $_POST['_nonce'], 'save_user_usergroups' ) ) {
 			die( __( 'Nonce check failed. Please ensure you can add users or user groups to a post.', 'edit-flow' ) );
 		}
 
 		$post_id = isset( $_POST['post_id'] ) ? (int) $_POST['post_id'] : 0;
-		$post = get_post( $post_id );
-		$user_group_ids = isset( $_POST['user_group_ids'] ) && is_array( $_POST['user_group_ids'] ) ? array_map( 'intval', $_POST['user_group_ids'] ) : array();
-		if ( isset( $_POST['ef_notifications_name'] ) && ! wp_is_post_revision( $post_id ) && ! wp_is_post_autosave( $post_id ) && current_user_can( $this->edit_post_subscriptions_cap ) ) {
-			if( $_POST['ef_notifications_name'] === 'ef-selected-users[]' ) {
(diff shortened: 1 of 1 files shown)


modules/notifications/notifications.php
```diff
@@ -360,25 +360,31 @@ function notifications_meta_box() {
...
+
+		$post_id = isset( $_POST['post_id'] ) ? (int) $_POST['post_id'] : 0;
+		$post    = get_post( $post_id );
+
+		$valid_post = ! is_null( $post ) && ! wp_is_post_revision( $post_id ) && ! wp_is_post_autosave( $post_id );
+		if ( ! isset( $_POST['ef_notifications_name'] ) || ! $valid_post || ! current_user_can( $this->edit_post_subscriptions_cap ) ) {
+			die();
+		}
+
+		$user_group_ids = isset( $_POST['user_group_ids'] ) && is_array( $_POST['user_group_ids'] ) ? array_map( 'intval', $_POST['user_group_ids'] ) : array();
```
sboisvert: We might want to break this line up or to add a comment as it's a bit long. We want to gravitate towards verbose easy to read code vs short code. The benchmark I try to use is, could I read this code if it was 2 am and something was broken that I had to fix. So in practice that means not using the ternary operator for anything that's not a simple check (Line 369 is a good usage of it). Of note, this is my personal preference and others might have more input on this :).

modules/notifications/notifications.php
```diff
@@ -360,25 +360,31 @@ function notifications_meta_box() {
...
 		
-		// Verify nonce
-		if ( !wp_verify_nonce( $_POST['_nonce'], 'save_user_usergroups') )
-			die( __( "Nonce check failed. Please ensure you can add users or user groups to a post.", 'edit-flow' ) );
-
-		$post_id = (int)$_POST['post_id'];
-		$post = get_post( $post_id );
-		$user_group_ids = is_array( $_POST['user_group_ids'] ) ? $_POST['user_group_ids'] : array();
-		$user_usergroup_ids = array_map( 'intval', $user_group_ids );
-		if( ( !wp_is_post_revision( $post_id ) && !wp_is_post_autosave( $post_id ) )  && current_user_can( $this->edit_post_subscriptions_cap ) ) {
```
sboisvert: I really like how you simplified the conditionals here.

modules/notifications/notifications.php
WPprodigy: I agree, it does err on the side of "too much" for one line / a ternary. Was the result of combining the previous two lines.

Issue #265, opened 2014-07-07
Shouldn't be able to subscribe someone to notifications if they can't edit post.


jerclarke:
Correct me if I'm wrong, but after some testing it seems that you are able to subscribe a user to a post in the Notifications box even if they don't have edit_others_posts and thus won't be able to "edit" the post and see Editorial Comments etc. 

My testing implies that this is at least partially corrected by just not sending notifications to users who wouldn't be able to edit. Is that right?

Either way it seems like a terrible idea to let people subscribe users who won't be able to log in and won't get notified. It's bound to cause confusion even among users that understand what's going on, but more than that it assumes that every user not only understands the roles+capabilities system in WP (i.e. that some users can/can't edit_others_posts, which takes our editors a while to grasp) but also knows what role each other user has (and thus whether they can successfully subscribe someone). 

If I'm not wrong above I think it would be much better if Edit Flow protected users from making the mistake of subscribing someone who won't get notified. 

One way would be to allow them to do so, but with an obvious warning that the user won't get any emails about the post because they can't edit. 

The other way would be to allow the user to be found while searching (to avoid confusion if they are missing) but de-activate the user from being subscribed with some kind of warning/label expressing the fact that they are inactive because they wouldn't have permission to edit the post. 

Hope that makes sense! Thanks for considering it :)

jeremyclarke:
Adding a quick note that I seem to have found a related bug: If such an author with no `edit_others_posts` capability is in a GROUP that is subscribed to a post, they DO get the email, but still can't log in. 

Very confusing, and seems to imply an unstated rule like **Never add users to groups unless they have edit_others_posts capability**, which, if it is effectively a constraint should be coded into the system with warnings etc. (and is also a terrible idea generally, since users can change roles and capabilities while they are members of a group, resulting in unpredictable outcomes).

sboisvert:
@WPprodigy I think we can start working on the notification part that's described in #273 and ties into this. 

>Adding a quick note that I seem to have found a related bug: If such an author with no edit_others_posts capability is in a GROUP that is subscribed to a post, they DO get the email, but still can't log in.

Seems like pretty bad UX :(

>Very confusing, and seems to imply an unstated rule like Never add users to groups unless they have edit_others_posts capability, which, if it is effectively a constraint should be coded into the system with warnings etc.

Seems like a good start, at least until we look into / fix #273

WPprodigy:
[...]

PR #449, opened 2018-04-04
Check if a user can edit the post/page before sending a notification.

WPprodigy:
Part of https://github.com/Automattic/Edit-Flow/issues/265

The functionality is all implemented in the first commit, the other two are cleanup for it's surroundings.

## Some things to note:

It's possible that the `user_can_be_notified` method will need to be abstracted and put into the `EF_Module` parent class for part 2 of the above issue, but that can be done if/when necessary.

And then it seems calling both `edit_post` or `edit_page` conditionally would be redundant according to the core method: https://github.com/WordPress/WordPress/blob/master/wp-includes/capabilities.php#L127-L128. Would be nice to have a second pair of eyes to confirm this, but my testing showed that this was indeed the case. Some more info on what I found here https://github.com/Automattic/Edit-Flow/issues/265#issuecomment-378419531

## Steps for testing

1) Add multiple authors and editors to a user group.
2) Using the current master branch, select this user group to be notified on a draft post. 
3) Post an editorial comment.
4) Look at the email logs or just error_log the `$recipients` variable at the end of `_get_notification_recipients`.
5) Note that all authors/editors that were signed up received this email.

5) Switch to this branch and post another editorial comment.
6) Note that only editors were sent the email, along with the post author if they were selected.

Can repeat and test with individual user notifications as well instead or alongside groups.

### DIFF
--- a/modules/notifications/notifications.php
+++ b/modules/notifications/notifications.php
@@ -757,14 +757,21 @@ jQuery(document).ready(function($) {
 				$usergroup = $edit_flow->user_groups->get_usergroup_by( 'id', $usergroup_id );
 				foreach( (array)$usergroup->user_ids as $user_id ) {
 					$usergroup_user = get_user_by( 'id', $user_id );
-					if ( $usergroup_user && is_user_member_of_blog( $user_id ) )
+					if ( $this->user_can_be_notified( $usergroup_user, $post_id ) ) {
 						$usergroup_users[] = $usergroup_user->user_email;
+					}
 				}
 			}
 		}
 		
 		$users = $this->get_following_users( $post_id, 'user_email' );
-		
+		foreach( (array) $users as $key => $user ) {
+			$user_object = get_user_by( 'email', $user );
+			if ( ! $this->user_can_be_notified( $user_object, $post_id ) ) {
+				unset( $users[ $key ] );
+			}
+		}
+
 		// Merge arrays and filter any duplicates
 		$recipients = array_merge( $authors, $admins, $users, $usergroup_users );
 		$recipients = array_unique( $recipients );
@@ -789,7 +796,27 @@ jQuery(document).ready(function($) {
 			return $recipients;
 		}
 	}
-	
+
+	/**
+	 * Check if a user can be notified.
+	 * This is based off of the ability to edit the post/page by default.
+	 * 
+	 * @since 0.8.3
+	 * @param WP_User $user
+	 * @param int $post_id
+	 * @return bool True if the user can be notified, false otherwise.
+	 */
+	function user_can_be_notified( $user, $post_id ) {
+		$can_be_notified = false;
+
+		if ( $user instanceof WP_User && is_user_member_of_blog( $user->ID ) && is_numeric( $post_id ) ) {
+			// The 'edit_post' cap check also covers the undocumented 'edit_page' cap.
+			$can_be_notified = $user->has_cap( 'edit_post', $post_id );
+		}
+
+		return apply_filters( 'ef_notification_user_can_be_notified', $can_be_notified, $user, $post_id );
+	}
+
 	/**
 	 * Set a user or users to follow a post
 	 *
--- a/modules/notifications/notifications.php
+++ b/modules/notifications/notifications.php
@@ -727,34 +727,35 @@ jQuery(document).ready(function($) {
 	function send_single_email( $to, $subject, $message, $message_headers = '' ) {
 		wp_mail( $to, $subject, $message, $message_headers );
 	}
-	
+
 	/**
-	 * Returns a list of recipients for a given post
+	 * Returns a list of recipients for a given post.
 	 *
-	 * @param $post object
-	 * @param $string bool Whether to return recipients as comma-delimited string or array 
-	 * @return string or array of recipients to receive notification 
+	 * @param WP_Post $post
+	 * @param bool $string Whether to return recipients as comma-delimited string or array.
+	 * @return string|array Recipients to receive notification.
 	 */
 	private function _get_notification_recipients( $post, $string = false ) {
 		global $edit_flow;
-		
+
 		$post_id = $post->ID;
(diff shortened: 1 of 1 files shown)


modules/notifications/notifications.php
```diff
@@ -727,69 +727,98 @@ function schedule_emails( $recipients, $subject, $message, $message_headers = ''
...
+	 * @return string|array Recipients to receive notification.
 	 */
 	private function _get_notification_recipients( $post, $string = false ) {
 		global $edit_flow;
-		
+
 		$post_id = $post->ID;
-		if( !$post_id ) return;
-		
-		$authors = array();
```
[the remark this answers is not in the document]
WPprodigy: `$authors` was never used. Was just set to an empty array, and then `array_merge`'d later on. The missing `$post_id` conditional is still present, just unfolded per WP coding standards.

modules/notifications/notifications.php
```diff
@@ -727,69 +727,98 @@ function schedule_emails( $recipients, $subject, $message, $message_headers = ''
...
-		
-		// Filter to allow further modification of recipients
+
+		// Merge arrays and filter any duplicates.
+		$recipients = array_unique( array_merge( $admins, $users, $usergroup_users ) );
+
+		// Process the recipients for this email to be sent.
+		foreach(  $recipients as $key => $user_email ) {
+			// Get rid of empty email entries.
+			if ( empty( $recipients[ $key ] ) ) {
```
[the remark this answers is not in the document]
WPprodigy: Hmm, good question. This check was legacy code that I left there, but re-reading looks like a possible typo. Should be checking if the email is empty, because it is technically possible to have a user without an email address.

Instead of the conditional, could just wrap `array_filter()` around the array manipulations above which will simply strip array any empty values when used without a callback function.

modules/notifications/notifications.php
```diff
@@ -727,69 +727,98 @@ function schedule_emails( $recipients, $subject, $message, $message_headers = ''
...
-		// Email all admins, if enabled
-		if( 'on' == $this->module->options->always_notify_admin )
+		// Email all admins if enabled.
+		$admins = array();
+		if ( 'on' === $this->module->options->always_notify_admin ) {
 			$admins[] = get_option('admin_email');
-		
+		}
+
 		$usergroup_users = array();
```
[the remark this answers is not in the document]
WPprodigy: Move `$usergroup_users = array();` to the top? I removed all of the top empty variable declarations. Some are no longer needed, and if they are, they are declared in a more localized place.

modules/notifications/notifications.php
```diff
@@ -727,69 +727,92 @@ function schedule_emails( $recipients, $subject, $message, $message_headers = ''
...
+	 */
+	function user_can_be_notified( $user, $post_id ) {
+		$can_be_notified = false;
+
+		if ( $user instanceof WP_User && is_user_member_of_blog( $user->ID ) && is_numeric( $post_id ) ) {
+			// The 'edit_post' cap check also covers the undocumented 'edit_page' cap.
+			$can_be_notified = $user->has_cap( 'edit_post', $post_id );
+		}
+
+		return apply_filters( 'ef_notification_user_can_be_notified', $can_be_notified, $user, $post_id );
```
david-binda: There is a catch with the `$user` as it might be boolean `false` under certain circumstances ( `get_user_by` call in `_get_notification_recipients` might return `false` ). Thus, I would propose to document the filter via PHPDoc notifying the users about that possible issue.

david-binda: Shall we cast the filtered value to `bool` in order to make sure that boolean is always returned from this method? Per PHPDoc comment?

modules/notifications/notifications.php
```diff
@@ -727,69 +727,92 @@ function schedule_emails( $recipients, $subject, $message, $message_headers = ''
...
-		foreach( $recipients as $key => $user_email ) {
-			// Get rid of empty email entries
-			if ( empty( $recipients[$key] ) )
-				unset( $recipients[$key] );
-			// Don't send the email to the current user unless we've explicitly indicated they should receive it
-			if ( false === apply_filters( 'ef_notification_email_current_user', false ) && wp_get_current_user()->user_email == $user_email )
-				unset( $recipients[$key] );
+
+		$user_recipients = $this->get_following_users( $post_id, 'user_email' );
+		foreach( (array) $user_recipients as $key => $user ) {
```
david-binda: Is the typecasting to `array` really necessary here? Looks like https://github.com/Automattic/Edit-Flow/blob/6fc7cb193c74a48581bdb0c5287ed35c7acbe763/modules/notifications/notifications.php#L956,L1000 is  always returning an `array`.

It might be worth digging into why it has been added to the original code in the first place. Or whether it was in there from the very beginning.

WPprodigy: I would be adding it in this PR, so you are right - it's not needed.

sboisvert:
This Looks good. I don't think we want to merge it into master yet but it's ready to go into a feature branch.

WPprodigy:
Agreed. I personally think https://github.com/Automattic/Edit-Flow/issues/451 needs to be ready before this goes into a release.

As without that piece, an editor could be mislead into thinking certain people are getting notifications when they are not.

rinatkhaziev:
@sboisvert do we want this to go into 0.8.3 or 0.9?

rinatkhaziev:
Per Slack - marking this for 0.9

jerclarke:
Tested this and it seems to work, aside from conflicts with the other patches as noted in the referenced links above. 

Would love if we could pick an official branch for all these changes and start merging them in. They should all be launched together in `0.9` and they have a lot of interactions that need to be addressed in a merged codebase (since they tend to interact with the same few files).

rinatkhaziev:
@WPprodigy thanks for the work, @jerclarke thanks for the testing. Merging that in to go 0.9 :)

PR #453, opened 2018-04-08
Display who will be notified of post status changes

WPprodigy:
Part of #270. This PR started in #272, and is continued with some edits here.

To test:

1) Enable Subscribers feature.
2) Subscribe some users and/or groups to a post.
3) Check the posts submit box to ensure it shows the correct counts and updates accordingly.

Screenshots can be found in #272, as there weren't many changes to the display.

### DIFF
--- a/modules/notifications/lib/notifications.css
+++ b/modules/notifications/lib/notifications.css
@@ -93,6 +93,6 @@
 	margin-top: 0;
 }
 
-.misc-pub-follower-count span {
+.misc-pub-follower-count #post-follower-count-display {
 	font-weight: bold;
 }
--- a/modules/notifications/lib/notifications.js
+++ b/modules/notifications/lib/notifications.js
@@ -58,27 +58,29 @@ var ef_displayFollowerCountInSubmitBox = function() {
 	var subscribed_users  = jQuery( '#ef-post_following_users_box li input:checkbox:checked' ).length;
 	var subscribed_groups = jQuery( '#ef-following_usergroups li input:checkbox:checked' ).length;
 
+	// Example: "1 user" or "23 users".
 	var users_message_part = '';
 	if ( subscribed_users > 0 ) {
 		if ( 1 === subscribed_users ) {
-			users_message_part = subscribed_users + ' user';
+			users_message_part = subscribed_users + ' ' + __ef_localize_notifications.user;
 		} else {
-			users_message_part = subscribed_users + ' users';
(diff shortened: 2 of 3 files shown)


WPprodigy:
Probably worth noting that the `following_list_updated` custom jQuery trigger is used/added in both PRs. So will need to deal with that merge conflict likely when it comes time to merge them both.

jerclarke:
FWIW after looking at the screenshots in #272 I am doubting the value of this change, at least in it's current form. 

Having the list take up space in the submitbox, just to list the count of users and groups, seems like a waste of time. 

I'd much rather have the full list of people who will be notified, as in the New Editorial Comment UI, as that lets you confirm your expectations about who is subscribed. Showing the counts leaves it easy to think someone is subscribed who isn't: 

image

Also, even when I consider having it show the full list of subscribers, it still seems weird, because we aren't saying what will happen. When will these people be notified? I know we know the answer, but users won't necessarily. By comparison, in the Editorial Comment box it's so explicit why the list is being shown: Submit this comment, these people get email. This submitbox use case is a lot weirder. 

Also FWIW I'd really rather avoid "Follower" as a new word for this context. Currently there's no use of that word in the post editor in the context of the **Notifications** or **Editorial Comments** metaboxes. The ticked users are just people who will be notified. 

Personally I think of them as "subscribers" rather than "followers", but as far as the post editor is concerned, neither of those words is part of the vocabulary yet. Follower somehow strikes me as wrong because people are automatically subscribed (when they edit) and subscribed by others (having their name ticked) so often. "Is following" feels like an active verb compared to "being subscribed" which describes the situation better. 

Either way, having "Followers" up in the top, without listing them, I think will be genuinely confusing. It sounds like a Twitter integration or something. If we had to label it with something, I can't help thinking it should say "Edit Flow Notifications: username, username, groupname" or something, which isn't great either. 

**TL;DR: As the person who requested this feature as part of #270, I realize now that it wasn't a great idea, and I can't think of a good UI for it, so we should just scrap it and instead enjoy the glory that is #452 's display of users who will be notified of an editorial comment.**

WPprodigy:
> I am doubting the value of this change, at least in it's current form.

> I'm not totally convinced this is necessary. If we want to keep it, I think we should show the full list of users.

I'm more-so leaning towards dropping this. I'm not sure it carries enough usefulness, and also there is GB to consider fairly soon. I think the Follower information should just be kept together and found in one place.

So +1 for passing on this. CC @sboisvert

rinatkhaziev:
Based on consensus in this discussion, closing this.

PR #456, opened 2018-04-18
Capabilities: add capabilities via filters on VIP Go sites.

jeherve:
In Edit Flow, we check for the `ef_view_story_budget` capability before to display some of the admin pages. The `ef_view_story_budget` capability should be given to admins, editors, and authors here:
- https://github.com/Automattic/Edit-Flow/blob/master/modules/story-budget/story-budget.php#L86-L101
- https://github.com/Automattic/Edit-Flow/blob/master/edit_flow.php#L194-L197

However, this does not get to run on VIP Go sites, thus leaving users without the custom caps necessary to access the different Edit Flow custom screens. To solve this, let's add all necessary caps via filters.

Reported in 76616-zd-wordpressvip

### DIFF
--- a/vipgo-helper.php
+++ b/vipgo-helper.php
@@ -5,6 +5,16 @@
 add_action( 'after_setup_theme', 'EditFlow' );
 
 /**
+ * Caps don't get loaded on install on VIP Go. Instead, let's add
+ * them via filters.
+ */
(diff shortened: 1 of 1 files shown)


rinatkhaziev:
Thanks, merging, will ship in 0.8.3

PR #459, opened 2018-04-23
dashboard-notepad.php: improved dashboard widget

mitchelldmiller:
Replaced get_the_time() with get_the_modified_time() to show correct time on footer.

Added post method to form tag to fix overflow of request line on long GET submission.

Use pagenow as current page value if wp_get_referer() returns false, to avoid blank page.


rinatkhaziev:
Thanks for the PR! 

Looks like it's a duplicate, and all of those issues were fixed at one point or another, and are a part of 0.8.3

PR #460, opened 2018-04-23
Story budget - Screen Option columns option fix

raduconst:
Made the changes discussed here: https://github.com/Automattic/Edit-Flow/pull/409#issuecomment-383417463

The old branch was no longer available so I created another PR. The only thing that I made different is that I added `columns-number-` in

```
echo '<div class="postbox-container columns-number-' . $this->num_columns . '">';
					foreach( (array) $this->terms as $term ) {
						$this->print_term( $term );
					}
```

because it wasn't working without this part. Basically, this is the initial number of columns that it's removed by this JS code:

```
$(".postbox-container").removeClass(function() {
				for (var index = 1, c = []; index <= columnsSwitch.length; index++) {
					c.push( classPrefix + index )
				}
				return c.join(' ');
			})
```

Everything else is just what you suggested.

### DIFF
--- a/modules/story-budget/lib/story-budget.css
+++ b/modules/story-budget/lib/story-budget.css
@@ -30,17 +30,16 @@
 	font-size: 14px;
 }
 
-
-.postbox-container {
-	float: left;
-	padding-right: 0;
-}
-
 .postbox {
 	display: block;
-	margin-bottom: 10px;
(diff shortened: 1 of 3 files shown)


rinatkhaziev:
Thank you, this is going into 0.8.3

Issue #451, opened 2018-04-05
Make it clear which users can/can't receive post notifications


WPprodigy:
Extracting the second part of #265 to a new issue. This comes after https://github.com/Automattic/Edit-Flow/pull/449 and ideally before/alongside https://github.com/Automattic/Edit-Flow/issues/273.

Only users that are able to view & edit a post will be receiving editorial comment and status change notification emails. But at the moment, users that do not have the `edit_post` privilege for specific post can still be signed up for notifications and the editor/author may not be aware that the person won't be receiving notifications.

In https://github.com/Automattic/Edit-Flow/issues/273, there will need to be a way to grant or at least show that a user is being given extra privileges by being selected for notifications. So this issue can start off the process with a basic implementation of just showing who can/can't receive notifications, then 273 can take it further with the needed functionality.

After some discussions, the consensus was that it would be best to do the permissions validation via ajax when the user is selected - rather than doing it for every user during page load. This is for scaling reasons, at least if/until we stop loading all users all at once.

So with that in mind, I'm thinking it should be something like this: http://cld.wthms.co/n3SUyc. This way, it only needs to run the permissions check when the user is being selected or if the user is already selected. Then the UI changes in 273 would simply be changing the label to "Access Granted" perhaps automatically or with the click of a button :)

![no-access](https://user-images.githubusercontent.com/8536129/38342485-3b79dcda-3844-11e8-90ce-0e153cab6f82.png)

Thoughts? @jerclarke @sboisvert

WPprodigy:
Also worth noting that this will also affect the two PRs in https://github.com/Automattic/Edit-Flow/issues/270.

The "who was/will be notified" messages should look for this "no access" badge and adjust accordingly.

Issue #273, opened 2014-08-11
Notifications module: Give author (no edit_others_posts) access to specific post via. 


jerclarke:
Related to #265 (don't allow subscribing users who won't be able to log into post due to insufficient permissions) this would solve the larger problem in a way that feels very natural to the goals of EditFlow.

The key goal is to allow an "editor" (someone with edit_others_posts capability who can thus log into any post and edit it) the ability to add an author (someone without edit_others_posts capability who can thus only edit posts where they are set as author) as an editor of a single post. This could be elegantly combined with the Notifications feature to solve #265 AND souping up the system as a whole. 
## Example workflow
- Someone editing a post and wants help from another user
- Finds user in Notifications metabox, but they are not an editor. 
- Notifications metabox makes it clear they can't be subscribed because they won't be able to see the post. 
- Notifications metabox instead offers a "Grant permission and subscribe"
- The "author" user is granted permission for the post and added to notifications, the post can now be treated as if they were an editor. 

I know that the roles/capabilities system has granularity down to the post_id in regards to the edit_posts capability, so it would just be a matter of storing a list of user_id's that a post can be edited by regardless of their capabilities (or a list of post_id's that a user can edit, but it seems much more scalable to do it by post rather than by user).
## Example Uses

This would be useful in many situations where a site uses the author/editor divide for editorial reasons, but ultimately trusts users to edit others posts in specific situations. 
- Several authors contributing to a post (but no multi-author plugin desired). 
- Summoning a topic expert without edit_others_posts to review the content of a post. 
- Allowing translators to edit specific posts they are translating. 

The specific uses are endless, but the key need is the ability to cooperate on posts across the newsroom without having to make almost all users into editors. Keeping the simple interface and non-publishing workflow of the 'author' role is important for many contributors, but fails to account for the nuance that applies to each post. 

I think this is a problem Edit Flow is in a prime position to solve because it will integrate so well with the Notifications feature. A theoretical external plugin would need to replicate pretty much the same UI to function, which would be terrible if both were running together. 

Thanks for considering it.

sboisvert:
@philipjohn @batmoo Do you have thoughts on if this functionality should be added? 
[...]

PR #461, opened 2018-05-04
Add a "No Access" badge to subscribers that will not be able to receive notifications

WPprodigy:
For https://github.com/Automattic/Edit-Flow/issues/451

Recap of what's being done in this PR:

1) Adds a "No Access" badge for users on page load. The permissions checks are only done for users that are already subscribed.
2) When users are selected and saved via ajax, the response sends back a list of checked user IDs that do not have access. Can then use this list to find out if the newly selected user needs the badge or not.

I prototyped a bit ahead, thinking of how #273 could be implemented. So some of the organization here had those changes in mind.

Also note that I added the `user_can_be_notified()` method in for testing purposes, but can remove/leave out when it comes to merging into the feature branch.

Screenshot:
no-access


jerclarke:
@rinatkhaziev Please see PR above, it's vital to get it in before any release, as this whole feature is currently broken without the fix.

Issue #361, opened 2016-08-31
Update styles on settings page 


cojennin:
The Edit Flow settings page could use some style updates so it doesn't look as funky on smaller screen sizes. See below screenshot for example.

![screen shot 2016-08-31 at 3 25 52 pm](https://cloud.githubusercontent.com/assets/1888152/18143249/3f248670-6f8f-11e6-9c2b-c86581b91f5c.png)

WPprodigy:
The above PR is merged and looks good in my testing 😀

Closing out this issue, thanks @TheCrowned

PR #464, opened 2018-05-28
Enlarged modules box up to 250px, so that buttons don't overlap anymore

TheCrowned:
Fix styling not being optimal on mobile devices (and small resolutions in general), as proposed in #361. 
In particular,
- Enlarged module boxes and button divs from 210px to 250px to prevent buttons from overlapping.
- Removed `min-width` property from modules container, which prevented the page from being fully adaptable to different resolutions.

Any viewer which is at least ~ 250px wide should see the page well. Tested with Chrome (with variable window size) and a mobile emulator. Now, if the viewer is not at least ~ 500px wide, module boxes are all stacked in a single column.

**Before CSS changes**

**After CSS changes**

### DIFF
--- a/modules/settings/lib/settings.css
+++ b/modules/settings/lib/settings.css
@@ -24,6 +24,8 @@
 	width: 100%;
 	overflow: hidden;
 	margin-top: 20px;
+	display: flex;
+	flex-wrap: wrap;
 }
 
 .edit-flow-modules .edit-flow-module {
@@ -89,6 +91,11 @@
 	box-shadow: none;
 	filter:  progid:DXImageTransform.Microsoft.gradient(startColorstr='#BBB', endColorstr='#888');
 	text-shadow: none;
+	float: right;
+}
+
+.edit-flow-modules .edit-flow-module .button-primary.enable-disable-edit-flow-module {
+	float: left;
 }
 
 .edit-flow-modules .edit-flow-module .button-primary.configure-edit-flow-module:hover {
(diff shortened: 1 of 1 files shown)


sboisvert:
@TheCrowned I'd add a bit more information to this PR's description, for example a screenshot of before and after on a mobile device (or a mobile emulator) While there is some more info in #361 it would be nice to have all the rationale for the PR here since Issues could have long conversations and not always start or go over the main reasoning in one place. Having a summary of the "why" in PRs is very useful (the "Why" can include things like screenshots to demonstrate).
It would also be nice to note what this change has been tested with and if it could have any unintended side effects.

TheCrowned:
All right - I have updated the PR initial description with details and screenshots. Also, I took the chance to make another commit further improving the layout on small screens (details above) :)

rinatkhaziev:
Thanks for your contribution @TheCrowned. Unfortunately, your solution doesn't really cover all the bases. We'll be better off with using Flexbox in conjunction with media queries. So that for smaller viewport a module should take 100% of widh and modules should be stacked on top of each other.

If you're unfamiliar with Flexbox, here's a good starting point: https://css-tricks.com/snippets/css/a-guide-to-flexbox/

I'm tentatively punting this to 0.9

TheCrowned:
Nice, I didn't know about the `flex` directives @rinatkhaziev! I have added them and they seem to work beautifully, even without the media queries. I have pushed an additional commit. I also took the chance to float the two buttons: one to the left and one to the right.

Tested the changes on my phone, see screenshots :)

PR #469, opened 2018-05-28
Notepad dashboard widget method changed to POST

### LINKED ISSUE #466: Notepad Dashboard widget, change form method to POST
Note data is sent as GET. Rejected by Apache when over 8192 bytes on my server.
Expected behavior: update note.
Actual behavior: Apache URL too long.
Steps to reproduce: save big note.

Extrapolated from #458.

TheCrowned:
Form method changed to POST and function `handle_notepad_update()` updated accordingly.
Fixes #466.

### DIFF
--- a/modules/dashboard/widgets/dashboard-notepad.php
+++ b/modules/dashboard/widgets/dashboard-notepad.php
@@ -59,9 +59,6 @@ class EF_Dashboard_Notepad_Widget {
 		} else {
 			wp_insert_post( $new_note );
 		}
-
-		wp_safe_redirect( wp_get_referer() );
-		exit;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


modules/dashboard/widgets/dashboard-notepad.php
```diff
@@ -59,9 +59,6 @@ public function handle_notepad_update() {
 		} else {
 			wp_insert_post( $new_note );
 		}
-
-		wp_safe_redirect( wp_get_referer() );
```
sboisvert: Is this change expected to be part of this PR?

modules/dashboard/widgets/dashboard-notepad.php
```diff
@@ -81,14 +78,14 @@ public function notepad_widget() {
 		$current_note = ( ! empty( $posts[0]->post_content ) ) ? $posts[0]->post_content : '';
 		$current_id = ( ! empty( $posts[0]->ID ) ) ? $posts[0]->ID : 0;
 		$current_post = ( ! empty( $posts[0] ) ) ? $posts[0] : false;
-
+
```
sboisvert: The whitespace should be part of a separate commit as per https://vip.wordpress.com/documentation/creating-good-changesets/

modules/dashboard/widgets/dashboard-notepad.php
```diff
@@ -112,4 +109,4 @@ public function notepad_widget() {
 		}
 	}
 
-}
\ No newline at end of file
+}
```
sboisvert: This change, while also fine, should probably not be part of this commit :)

modules/dashboard/widgets/dashboard-notepad.php
TheCrowned: Yes. When data was sent over GET, that line had the use of hiding the querystring. The user was immediately redirected to the page it created the note from, so that he would not see the GET parameters. However, [wp_get_referer returns false after a POST request](https://codex.wordpress.org/Function_Reference/wp_get_referer), so I had to be taken away, together with the following `exit`.

TheCrowned:
I have removed the whitespace changes :)

I will create another branch + PR to change $_REQUEST usage in $_POST usage in this file.

sboisvert:
cc @rinatkhaziev to decide in what release to stick this in

PR #470, opened 2018-05-30
Changed $_REQUEST usage to $_POST usage in handle_notepad_update()

TheCrowned:
The function `handle_notepad_update()` used to use `$_REQUEST`, which is [against best practices](https://vip.wordpress.com/documentation/code-review-what-we-look-for/#using-_request). 

Before #469, the form did not specify a method so it defaulted to GET. With #469, we changed the method to POST, so all references to `$_REQUEST` have been updated with the corresponding `$_POST` ones.

**NOTE** that this depends on #469, because it expects the HTML form to use the POST method, or the note saving will just not work.

### DIFF
--- a/modules/dashboard/widgets/dashboard-notepad.php
+++ b/modules/dashboard/widgets/dashboard-notepad.php
@@ -35,7 +35,7 @@ class EF_Dashboard_Notepad_Widget {
 		global $pagenow;
 
 		if ( 'index.php' != $pagenow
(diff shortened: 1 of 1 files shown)


sboisvert:
Looks good, great commit message!

One thing that we could add is that it might be useful in the note to state what commit / PR # this depends on. (It's can be deduced from the rest of the text that it depends on #469 but it could be useful to have really stand out so folks skimming it quickly still catch it.

Issue #478, opened 2018-11-05
Notifications: Include list of people who were notified in the editorial comment emails


jerclarke:
One of our editors wants to be able to add a Gmail filter to add tags to EF email based on which group was ticked in the notifications box. This seems reasonable, but right now there's no way to know why you got a given notification, it just comes to the email of your WP account. 

If you've got a complicated EF+WP+email setup, some people are liable to get many EF emails for a single post change, and not knowing why the emails are coming can be confusing. 

A related issue is that it would be nice to know who else got the notification from right in the email, e.g. so you know that the correct person saw it and you can trust them to handle the situation for now. This is basically the "other side" of #270 , that one is about knowing who gets the notification while sending it, this one is about knowing who else received it. 

I see two main solutions that address these related goals:

### Option 1:  Indicate the list of subscribers in all emails

```
___________
This notification was delivered to:

Group: Sub-Editors
Group: Time-Sensitive
Jer Clarke
Adam Adamson
Susan Susanson
```

###  Option 2: Indicate in individual emails why it is being sent to that address

```
__________
You are subscribed to this post as: Jer Clarke
You were subscribed to this post via the group(s): Sub-Editors, Time-Sensitive
```

Either of these would solve the precise use-case of my user (adding a Gmail filer) and generally enable users to put together why they got an email, since at least for the use cases I'm imagining, seeing the full list of subscribers would usually let someone figure out how it ended up in their inbox. 

The second option has a theoretical privacy benefit, as we don't leak the full list of subscribers, but that's of limited value because the user can presumably log in and see the same information on the post itself. Also the "display names" of fellow subscribers will rarely be a security risk, though it's possible. Being able to disable the `show_other_subscribers` feature via. a filter should be enough for the few sites security-conscious enough to consider this a problem. 

"Option 2" has the precise benefit of making it totally explicit why this exact email is coming to you, which could have utility e.g. debugging why you get multiple copies of the same email (multiple WP accounts). That said, I suspect that personalizing each email would be more coding and complexity, in which case IMHO "Option 1" is a very nice compromise that would just involve generating a list of subscribers for the post and adding it to the generic template that goes out to everyone. 

Of course there's also an "Option 3" that would clearly be the most work to execute but maybe not that much more work than options 1 or 2: 

###  Option 3: Indicate in individual emails why it is being sent to that address AND include a list of other subscribers

```
__________
You are subscribed to this post as: **Jer Clarke**
[...]

PR #479, opened 2018-11-06
Update EF_Notifications->notification_comment() to insert the list of users/groups who were notified

jerclarke:
For #478  

- Updates phpdoc to specify that $comment is WP_Comment
- Near the top, we fetch the list of users/groups from comment meta
- Before the "actions" list, we output the list of people who were notified

Note: This template could use so much love, but for now focusing on putting this new feature in a sensible place, with a plan to return and redesign the template later.

### DIFF
--- a/modules/notifications/notifications.php
+++ b/modules/notifications/notifications.php
@@ -658,12 +658,12 @@ jQuery(document).ready(function($) {
 		}
 		*/
 		
+		
+		$body .= "\r\n--------------------\r\n";
 		// Insert the notification list from comment meta @see EF_Editorial_Comments->maybe_output_comment_meta()
 		if ($notification_list) {
 			$body .= esc_html__( 'Notified', 'edit-flow' ) . ": $notification_list\n";
 		}
-		
-		$body .= "\r\n--------------------\r\n";
 		
 		$edit_link = htmlspecialchars_decode( get_edit_post_link( $post_id ) );
 		$view_link = htmlspecialchars_decode( get_permalink( $post_id ) );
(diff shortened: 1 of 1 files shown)


jerclarke:
Okay I just added another commit to my branch, and it's showing above, so I guess it's part of this PR now which is good. 

It just moves the message down below the "---------" line which makes it clearer. 

Here's a screenshot of how it looks with latest version:

ef-editorial-comment-email-with-notified-list

modules/notifications/notifications.php
```diff
@@ -602,6 +602,9 @@ function notification_status_change( $new_status, $old_status, $post ) {
 	
 	/**
 	 * Set up and set editorial comment notification email
+	 * 
+	 * @param WP_Comment $comment
+	 * @return boolean
```
WPprodigy:
```suggestion
	 * @return boolean|null|void
```

WPprodigy: Could probably just make both of the early `returns` return the same thing to simplify this, either null for both or false for both. Not a big deal though.

modules/notifications/notifications.php
```diff
@@ -652,7 +658,12 @@ function notification_comment( $comment ) {
 		}
 		*/
 		
+		
 		$body .= "\r\n--------------------\r\n";
+		// Insert the notification list from comment meta @see EF_Editorial_Comments->maybe_output_comment_meta()
+		if ($notification_list) {
+			$body .= esc_html__( 'Notified', 'edit-flow' ) . ": $notification_list\n";
```
WPprodigy:
```suggestion
			$body .= esc_html__( 'Notified', 'edit-flow' ) . ": " . esc_html( $notification_list ) . "\n";
```

WPprodigy: Need to escape the meta output (untested regarding the string concatenation)

jerclarke:
Tested your changes and everything seems to work, thanks Caleb.

rinatkhaziev:
Thanks, merging in 0.9

Issue #480, opened 2018-11-06
Make "activate module" buttons default button-primary blue (remove green)


jerclarke:
I've been ignoring these ugly buttons (blue shading on the text) for years:

<img width="501" alt="ef-buttons-greenblue" src="https://user-images.githubusercontent.com/139995/48099172-5a628800-e1e4-11e8-9005-bd67badc7198.png">

Now it would be easy enough to fix the shading on the text to not be there, but it poses an existential question: **Why are these buttons green at all?**

There's nothing green about editflow.org, it's theme is that dark red. AFAICT it's just these buttons. 

I propose they would work perfectly as the default `button-primary`-blue. 

---

Sorry, meant to do a PR immediately, but ran into some weird Git behavior I need to work out. Will submit one shortly,

PR #481, opened 2018-11-08
settings.css - remove green color on module-enabling button-primary buttons

jerclarke:
For #480 

The green color doesn't make a lot of sense, and has had an ugly blue shadow for a long time.

Why not remove the green entirely? It looks good with blue buttons, and won't cause headaches in the future if button-primary changes again.


rinatkhaziev:
Thanks for this one, merging. Closes #480

PR #490, opened 2018-12-22
Wrap network check in function_exists() block

### LINKED ISSUE #255: Fatal error: Call to undefined function get_current_screen()
This error occurred when I was trying to update email information from a woocommerce my-account/edit-account screen as both admin and as a customer which is on the front-end.

Fatal error: Call to undefined function get_current_screen() in edit-flow/modules/user-groups/user-groups.php on line 693

Fix was made by adding to line 11 at the top: 
require_once (ABSPATH . wp-admin/includes/screen.php);

The error is not showing and appears to be fixed.

My setup is not a network site, but has multiple vendors set up, so I do not know if this is a problem with normal woocommerce sites and the use of this plugin.

### LINKED ISSUE #489: Front-end profile update causes Fatal Error.
When updating a user profile in Ninja Shop, formerly Exchange, Edit Flow throws:

> Fatal Error: Uncaught Error: Call to undefined function get_current_screen()

While a duplicate of #255, the issue was not resolved by #283.

kjohnson:
Resolves #489.

Resolves #255.

This pull request adds a check for `is_admin` before checking if the `get_current_screen()` function exists.

The function `get_current_screen()` "[is defined on most admin pages, but not all.](https://codex.wordpress.org/Function_Reference/get_current_screen#Usage_Restrictions)". This suggests that the function should be checked before being called, particulalry if not the code is used on admin and non-admin pages alike.

Specifically, this code runs on the `user_profile_update_errors` action which is triggered by front-end user profile pages.

### DIFF
--- a/modules/user-groups/user-groups.php
+++ b/modules/user-groups/user-groups.php
@@ -687,9 +687,13 @@ class EF_User_Groups extends EF_Module {
 		if ( !$update )
 			return array( &$errors, $update, &$user );
 
-		//Don't allow update of user groups from network
-		if ( ( !is_null( get_current_screen() ) ) && ( get_current_screen()->is_network ) )
-			return;
+		// `get_current_screen()` is defined on most admin pages, but not all.
+		if( is_admin() && function_exists( 'get_current_screen' ) ){
+			//Don't allow update of user groups from network
+			if ( ( !is_null( get_current_screen() ) ) && ( get_current_screen()->is_network ) ) {
+				return;
+			}
+		}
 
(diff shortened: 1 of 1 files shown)


rinatkhaziev:
@kjohnson I'm not sure I understand the check for `is_admin`, it's true that global `$wp_screen` is not defined until after `current_screen`action but the class `WP_Screen` and all screen-related functions get loaded on each admin request (including admin-ajax.php). (see wp-admin/includes/admin.php), so `get_current_screen` will not be defined on frontend ever, and the check for `is_admin` is redundant.

kjohnson:
@rinatkhaziev sure, the check for `is_admin()` is redundent - I had added it to provide context for a call to `get_current_screen()`, which is specific to the admin.

That said, the `is_admin()` check can be removed.

dchymko:
Tested to confirm, going to update this PR to merge into a branch on the upstream before merging to master

Issue #483, opened 2018-11-09
Editorial comments "Notified" list should only include people who get email (not current_user)


jerclarke:
This is a follow-up for #270 (show who is getting notified of comments) which is largely complete and the initial version has been merged to master, so I'm making a new ticket. This issue also directly affects #478  (show the notified list in the email notifications too) since that code uses the same list of users that gets saved into the `notification_list` comment meta based on the JS-driven metabox UI. 

**MAIN SYMPTOM: When generating the list labeled "The following will be notified" in the Editorial Comments UI, the person editing the post is included in the list, even though, based on all the logic in this plugin, they are of course not notified.**

I noticed this while testing #449 ("Check if a user can edit the post/page before sending a notification.") because I was comparing the list of people in the UI and the total emails sent, and of course, my name was listed in the UI but I wasn't sent an email. 

**WIDER ISSUE: The list has everyone subscribed, rather than only those who are sent an email.**

I didn't notice this problem with the "current user" being listed until now (and presumably others didn't) because it is very natural to see your own user account listed in the UI. Of course as someone editing the post you are "subscribed" in **Notifications**, and in most ways the list displayed as "The following will be notified" is just a list of all subscribers. 

But that's not what it *should* be. It should be a list of only those people who will, in fact, be notified. And when it's displayed later -- when hovering on the comment and in the email notification -- it should really reflect exactly who got a notification (i.e. which emails were sent out). 

The use case I've already brought up, of the "current_user" being listed, is pretty low-priority, since it's unlikely to cause harm and, obviously, whoever wrote the comment "notified themselves" as they wrote and submitted the comment. 

**The important part is the "subscribed" people who don't get notified for various reasons.**

This gets a lot more important when you take into account the conjunction of #265 (original parent ticket) #451 (@WPprodigy 's more specific ticket) and #461 (@WPprodigy 's PR for #451). All three of these relate to how we handle the subscription of users who aren't able to edit a post because they don't have the `edit_others_posts` capability (i.e. "NO ACCESS"). 

Essentially, we need to take those people into account in the following ways (this is the stuff in the other tickets):

- Warning users when they subscribe someone with NO ACCESS
- Not sending the notification to someone with NO ACCESS
- A system for granting access to this specific post to users who otherwise have NO ACCESS

Considering all that, I think we can see how there is a very real difference between "Someone subscribed in the notifications box" and "Someone who will be notified". 

[...]

PR #492, opened 2019-01-09
For #483 - Updates notifications js/php to only show users who will be notified

jerclarke:
### I've now added all three major changes needed to ensure only users who will be sent a notification are listed in the Editorial Comments "Notification list". 

No Access and No Email badges both work using a unified system now, and those users aren't added to the "Notified list". 

The current user's checkbox is marked with a special class, and the "Notification list" JS is set up to ignore the current user. 

### Text below describes the original commit.

For #483 - Updates notifications js/php to avoid "No Access" users showing in the notified list in editorial comments

Adds `.post_following_list-no_access` to the "No Access" warning spans (in two places), then uses that during "Notified list" generation to exclude people without access from the list (which already happens during email sending, this just cleans up the display so it matches the actual outcome).

My intention is to add similar tests for current_user and no_email users.

- `notifications.php` - add `.post_following_list-no_access` to the initial PHP-created "No Access" span
- `notifications.js` - add `.post_following_list-no_access` to the post-AJAX JS-created "No Access" span
- `notifications.js` - Move the triggering of `following_list_updated()` to AFTER the no access stuff runs, so that it can see the span. Previously the "notified list" was updated before, so it couldn't see any of the results of the access test
- `editorial-comments.js` - before inserting a user into the "notified list" array, check if a span indicating a problem is next to the checkbox

### DIFF
--- a/common/php/class-module.php
+++ b/common/php/class-module.php
@@ -469,13 +469,16 @@ class EF_Module {
 
 		<?php if( !empty($users) ) : ?>
 			<ul class="<?php echo esc_attr( $list_class ) ?>">
-				<?php foreach( $users as $user ) : ?>
-					<?php $checked = ( in_array($user->ID, $selected) ) ? 'checked="checked"' : ''; ?>
+				<?php foreach( $users as $user ) : 
+					$checked = ( in_array($user->ID, $selected) ) ? 'checked="checked"' : '';
+					// Add a class to checkbox of current user so we know not to add them in notified list during notifiedMessage() js function
+					$current_user_class = ( get_current_user_id() == $user->ID ) ? 'class="post_following_list-current_user" ' : '';
+				?>
 					<li>
 						<label for="<?php echo esc_attr( $input_id .'-'. $user->ID ) ?>">
 							<div class="ef-user-subscribe-actions">
 								<?php do_action( 'ef_user_subscribe_actions', $user->ID, $checked ) ?>
-								<input type="checkbox" id="<?php echo esc_attr( $input_id .'-'. $user->ID ) ?>" name="<?php echo esc_attr( $input_id ) ?>[]" value="<?php echo esc_attr( $user->ID ); ?>" <?php echo $checked; ?> />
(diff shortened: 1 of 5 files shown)


jerclarke:
Third commit is the final fix for #483, fixing the "notified list". 

This change avoids having the current user displayed as a recipient because they won't be notified (they are sending the message after all).

class-module.php
- add a CSS class to the checkbox for the "current user" so we can check it later and not show yourself in notified list

editorila-comments.js
- during notifiedMessage() skip users who's checkbox has the class

rinatkhaziev:
@jerclarke This looks great, I only have a minor note, can we make the colors of badges match the official colors: https://make.wordpress.org/design/handbook/design-guide/foundations/colors/#grays or https://make.wordpress.org/design/handbook/design-guide/foundations/colors/#hues

I'd personally pick one of grays, but I defer to you.

This is somewhat important because as it is right now, the plugin is full of arbitrary colors and a bit off-brand, and I'd like to change that situation or at the very least not to aggravate it :)

jerclarke:
@rinatkhaziev I would love to! The color there now wasn't my choice and I was going to propose a follow up change to make it higher contrast, that faint red doesn't have enough contrast to pass any kind of accessibility test. 

Will do a patch right ow

jerclarke:
screen shot 2019-01-10 at 11 03 23 am

Here's how it looks after the patch. Ugly but appropriate, since it's important that they notice the warnings, and the amount of contrast is much more accessible!

Sidenote: I think we should also change the "green flash" when a subscription is saved (upon AJAX return) to be a "red flash" when there's a warning, but haven't looked into what needs to be changed for that yet.

rinatkhaziev:
@jerclarke 

Looking good! Merging and then I'll be spending the rest of the day doing chores like updating the changelog and readme, but this definitely goes out today. 

Thanks for your work both on code and planning/raising concerns.

jerclarke:
Thanks! So glad this new version will be out there!

PR #493, opened 2019-01-10
notifications.js - Update background color that flashes when ticking/unticking users

jerclarke:
( @rinatkhaziev I know you said you were trying to wrap up for today, but it turned out pretty easy to make this change and I think it goes well with the others. The colors are from the WP colors codepen)

Before: It would flash green as long as the AJAX came back
Now: It flashes red when you've ticked a user with a warning, otherwise green

- sets a `warning_background = true` flag during `toggle_warning_badges()`
- uses that flag in the AJAX response to choose between green and red background
- re-organizes the code a bit to run `toggle_warning_badges()` before the animation and fix an indentation issue


rinatkhaziev:
@jerclarke thanks, merging!

PR #499, opened 2019-01-26
Replace “$” with “jQuery”.

jameslesliemiller:
Was getting a missing function error for the $, and it was creating a conflict with another plugin. As I understand it, when using jQuery in WordPress, best practice is to always use the jQuery function instead of the $. Updated this file (the source of the error) and everything is peachy. It would be great to have this change rolled into the plug-in, so that I don’t have to keep making it after each update.

Also, I hope I’m going about this the correct way. I got followed the instructions from the plugin website (I think). I’ve never use git for collaboration, so if I did this wrong, please feel free to correct me, it won’t hurt my feelings at all.

### DIFF
--- a/modules/editorial-comments/lib/editorial-comments.js
+++ b/modules/editorial-comments/lib/editorial-comments.js
@@ -23,7 +23,7 @@ editorialCommentReply = {
 		jQuery('a.ef-replysave', row).click(function() { return editorialCommentReply.send(); });
 
 		// Watch for changes to the subscribed users.
-		$( '#ef-post_following_box' ).on( 'following_list_updated', function() {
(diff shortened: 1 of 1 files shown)


jameslesliemiller:
It would be great if someone would merge this PR. It’s been sitting here for 4 months, and it fixes an issue that’s breaking functionality in other plugins.

jameslesliemiller:
Please, somebody look at this PR and merge it.

jameslesliemiller:
Is anybody even maintaining this plugin anymore?

szepeviktor:
@rebeccahum Could you Merge this one please?

jameslesliemiller:
THANK YOU!

PR #500, opened 2019-02-21
Fix “$ is not a function” like error in panel

anboto:
There is a syntax error.


anboto:
Closes as duplicate of #499

PR #502, opened 2019-03-08
notifications.php - fix old_status_friendly_name undefined warning

jerclarke:
- affects `notification_status_change()`
- Moves the definition of $old_status_friendly_name (as well as $new_status_friendly_name) to above the complex `else if` block to ensure that it is always defined.
- Fixes a recent change to this code, which moved the definition of $old_status_friendly_name to places where it didn’t effect all of the `else if` clauses:

https://github.com/Automattic/Edit-Flow/commit/071791d0ed2ec676895f4f133e16642a54b17c65#diff-59ee072dc425b09b6d748cb84634734b

- This triggered a `PHP Warning: PHP Notice:  Undefined variable: old_status_friendly_name` because the variable is used right after in all cases.
- This warning also reflected that the string would be missing in many cases.
- I tried to test this with all the possible status transition but might have missed some. Certainly the common ones of new->draft->pending->publish->trash are all working as expected now.

### DIFF
--- a/modules/notifications/notifications.php
+++ b/modules/notifications/notifications.php
@@ -583,6 +583,12 @@ jQuery(document).ready(function($) {
 				$current_user_email = '';
 			}
 			
+			// Set up default old and new status "friendly" names using it's label
+			$old_status_post_obj = get_post_status_object( $old_status );
+			$old_status_friendly_name = $old_status_post_obj->label;	
+			$new_status_post_obj = get_post_status_object( $new_status );
+			$new_status_friendly_name = $new_status_post_obj->label;
+			
 			// Email subject and first line of body 
 			// Set message subjects according to what action is being taken on the Post	
 			if ( $old_status == 'new' || $old_status == 'auto-draft' ) {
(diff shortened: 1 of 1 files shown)


cojennin:
I'd like to merge this in, but it reintroduces a problem mentioned [here](https://github.com/Automattic/Edit-Flow/pull/446#discussion_r177689103). 

On way it triggers: If you programmatically call `wp_insert_post` then the `$old_status` will be `new` and `get_post_status_object` will return null, and you'll get an exception [here](https://github.com/Automattic/Edit-Flow/pull/502/files#diff-59ee072dc425b09b6d748cb84634734bR588).

One thing I can think of is adding null check around `$old_status_post_obj` and `$new_status_post_obj` and make `$old_status_friendly_name` and `$new_status_friendly_name` default to something, maybe an empty string? 

Open to suggestions. `=> New Status` doesn't seem that odd if we're missing the old status, but `Old Status =>` might be confusing if we're missing the new status

mjangda:
Continued in #560 560

PR #503, opened 2019-03-17
Fixed the php 7.2 count warning on the Settings page

raduconst:
`$edit_flow->modules` is an object with all the active modules but `count` cannot be used on it because it's not a countable object. 
The `Settings` page where this warning appears is a module too so this check is redundant since the result here will always be at least 1.
However, in order to make the warning go away, I choose to  use `get_object_vars` because it's turning `$edit_flow->modules` object into an array of objects (that can be counted).


mjangda:
Replaced by #534

Issue #25, opened 2012-01-27
Reinstate "Toggle Post Details" functionality on the story budget


danielbachhuber:
We've temporarily removed the excerpt and the toggling functionality

GaryJones:
This was implemented in PR #846 with the following approach:

**Implementation:**
- Added a "Show post excerpts" checkbox in Screen Options
- Post excerpts are displayed below the post title in the Story Budget table
- The setting toggles instantly without page reload via JavaScript
- User preference is saved to user meta and persists across sessions

**Excerpt logic** (same as the original PR #504 approach):
1. Password-protected posts show "There is no excerpt because this is a protected post."
2. Posts with a custom excerpt display that excerpt
3. Posts without an excerpt use `wp_trim_words()` to generate one from post content (default 20 words)

**CSS/JS:**
- Excerpts are wrapped in `.post-details` div, hidden by default
- When enabled, `.show-excerpts` class is added to the wrapper, making excerpts visible
- Screen Options checkbox triggers instant toggle via JavaScript class manipulation

PR #504, opened 2019-03-27
Added back the Toggle Post Details option to the Story Budget page

raduconst:
This PR is a fix for #25 
The `Toggle Post Details` button was removed here https://github.com/Automattic/Edit-Flow/commit/95c6874be7a9f00abcaa09811ce11104c7d99eb4#diff-9163f1310b94e41b397251a77e18ec5a but the `setup_postadata` is still not working so I couldn't use the `the_excerpt` function. 

What I did instead is try to replicate the functionality that's used on the <em>Dashboard → Pages</em> area and take care of these three cases:
1. when the post is password protected, the `There is no excerpt because this is a protected post.` message is displayed (ref https://core.trac.wordpress.org/browser/tags/5.1.1/src/wp-includes/post-template.php#L400);
2. when there's an excerpt set by the user, display it (ref https://core.trac.wordpress.org/browser/tags/5.1.1/src/wp-includes/post-template.php#L412);
3. when there's no excerpt set, use `wp_trim_words` like explained here: https://developer.wordpress.org/reference/functions/get_the_excerpt/#comment-content-2457 Basically, this will generate automatically a text from the post content (ref. https://core.trac.wordpress.org/browser/tags/5.1.1/src/wp-includes/formatting.php#L3713).

I also made a little change on L474 and add the `post-title` class to the title column since the JS code that's adding the `hidden` class is still available.

### DIFF
--- a/modules/story-budget/story-budget.php
+++ b/modules/story-budget/story-budget.php
@@ -471,7 +471,11 @@ class EF_Story_Budget extends EF_Module {
 		?>
 		<tr id='post-<?php echo esc_attr( $post->ID ); ?>' class='alternate' valign="top">
 			<?php foreach( (array)$this->term_columns as $key => $name ) {
-				echo '<td>';
+				if ( $key === 'title' ) {
+					echo '<td class="post-title">';
+				} else {
+					echo '<td>';
+				}
 				if ( method_exists( $this, 'term_column_' . $key ) ) {
 					$method = 'term_column_' . $key;
 					echo $this->$method( $post, $parent_term );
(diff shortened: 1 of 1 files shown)


GaryJones:
Closing in favor of PR #846, which has been merged.

The Toggle Post Details functionality from this PR was incorporated into #846 with the following enhancements:

- Moved the toggle to Screen Options as a "Show post excerpts" checkbox
- Added instant toggle via JavaScript (no page reload needed)  
- User preference saved to user meta and persists across sessions
- Same excerpt logic: handles password-protected posts, custom excerpts, and auto-generated excerpts

Thank you @raduconst for the original implementation approach - the excerpt generation logic was used as the foundation for this feature.

PR #506, opened 2019-04-12
editorial-comments.js - replace faulty `$()` with `jQuery()` to match the rest of file

jerclarke:
- On one of my sites, I was getting a site-breaking js error for `$ not defined` caused by line 26 of this file, which had a `$()` jQuery command
- For some reason, most of my sites never had this problem, but this one did. Maybe it's a combination of plugins and a common one makes `$()` available by coincidence.
- Either way, the surrounding code makes it obvious that `editorial-comments` relies on `jQuery()` being available, rather than `$()`
- This patch just fixes that one line (which was the most recently added, it's part of the "notified list" feature added in 0.9) to match the rest and use `jQuery()` like a good kid.

### DIFF
--- a/modules/editorial-comments/lib/editorial-comments.js
+++ b/modules/editorial-comments/lib/editorial-comments.js
@@ -23,7 +23,7 @@ editorialCommentReply = {
 		jQuery('a.ef-replysave', row).click(function() { return editorialCommentReply.send(); });
 
 		// Watch for changes to the subscribed users.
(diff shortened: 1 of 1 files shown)


szepeviktor:
@jacklenox #499 is already there.

szepeviktor:
#499 was merged!

dchymko:
Closing as #499 addressed this.

PR #507, opened 2019-04-12
editorial-comments.js - fix “no one will be notified” message logic

jerclarke:
Note: Noticed this while working on #506 and they work on the exact same area of the code, so it may be tricky to merge them both. You could easily update this patch to include the changes from #506 if it makes it easier for you.

### PROBLEM:

- When no one will be notified of an Editorial Comment the "No one will be notified" message should of course show in the space where there would otherwise be a list of usernames
- Right now, most of the time, the message is missing, and NOTHING shows
- The reason is that the message currently only shows when exactly zero users are subscribed, but the post author is ALWAYS subscribed. So if the post author is writing a comment when they are the only subscriber the message doesn't show, leaving an empty space.
- This is an obvious bug. If there’s no usernames to show, then the “no one will be notified” message should show instead.

### SPECIFIC CAUSE IN THE CODE: 

The JS checks for “no subscribers” BEFORE removing the one’s who won’t be notified (due to being the comment author, not having permission to edit the post, or not having an email), so the current author's presence on the list stops the message from being shown, then it doesn’t check again after running the filter and removing current author.

### SOLUTION:

- In the JS, we run our logic that removes people from the list if they won’t be notified BEFORE  we check if there are subscribers (and exit if there are none).

### NOTES:

- I’m very embarrassed I didn’t notice this before, but testing these things gets very abstract, and apparently no one else spotted it either 🤷🏻‍♀️


WPprodigy:
Ah, makes sense - nice catch. I tested and this does resolve the issue.

jerclarke:
FWIW I just tested this with 0.9.6 and it still works and is still a really obvious way to make that field work better.

jerclarke:
FWIW if anyone is still working on EF, this is still a great thing to fix!

jerclarke:
This is still relevant with the latest version, would love to see the fix integrated. Who would be harmed? Just more clarity when leaving a comment!

jerclarke:
Updated this for the latest EF version, would be really great to see this integrated.

GaryJones:
(Toggled status to get the CI to re-run.)

GaryJones:
Hi @jerclarke, thanks for this fix! The logic change makes sense - moving the "no one will be notified" check to *after* the filtering loop is the right approach.

However, I think there might be a small issue with the implementation. After moving the check, it still checks `subscribed_users.length`:

```javascript
if ( 0 === subscribed_users.length ) {
```

But `subscribed_users` is the original unfiltered list. Shouldn't this now check `usernames.length` instead, since that's the filtered list after removing users who won't be notified?

```javascript
if ( 0 === usernames.length ) {
```

Otherwise, the check would still pass if there are subscribed users, even if they've all been filtered out (e.g., the only subscriber is the current user writing the comment).

Does that match your understanding, or am I missing something?

jerclarke:
Gary, you are absolutely right. Sorry about that, I'm not sure why the PR was wrong but my hotfix I've been running indeed uses `usernames.length`:

```php
if ( 0 === usernames.length ) {
	message_wrapper.addClass( 'ef-none-selected' );
	message_wrapper.val( __ef_localize_post_comment.none_notified );
	return;
}
```

Do you need me to update the PR, or can you fix it on your end?

GaryJones:
If you want to do it @jerclarke, then I'm not hijacking your work 😄 

Thank you!

jerclarke:
Ok hopefully this is correct! Sorry my experience with updating PRs is a work in progress, so hopefully what's there is acceptable 🙏🏻

PR #509, opened 2019-04-15
Added a button to the Story Budget date change option

### LINKED ISSUE #176: Add a "Today" button to the Story Budget
If your current date range is far in the past, it's a little time consuming to jump to today. We should make that easy to do with a button similar to what we have on the calendar.

raduconst:
I added the button to let the user go faster to the current day instead of having to use the
calendar. I tried to keep the same behavior of the Today button from Edit Flow → Calendar → Customizable calendar, where you can go to the current date with just a press of a button. 

P.S. If this PR is accepted, the pot file needs to be generated again because there's a `or` word between the `Change` and `Today` buttons.

This is made to fix #176

### DIFF
--- a/modules/story-budget/story-budget.php
+++ b/modules/story-budget/story-budget.php
@@ -194,8 +194,10 @@ class EF_Story_Budget extends EF_Module {
 	 * @since 0.7
 	 */
 	public function handle_form_date_range_change() {
+		$is_today_submit = isset( $_POST['ef-story-budget-today-submit'] );
+		$is_range_submit = isset( $_POST['ef-story-budget-range-submit'], $_POST['ef-story-budget-number-days'], $_POST['ef-story-budget-start-date_hidden'] );
 
-		if ( ! isset( $_POST['ef-story-budget-range-submit'], $_POST['ef-story-budget-number-days'], $_POST['ef-story-budget-start-date_hidden'] ) ) {
+		if ( ! $is_today_submit && ! $is_range_submit ) {
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


dchymko:
Thanks for adding this!  

I made a minor change to relocate the Today button beside the date field for better context

modules/story-budget/story-budget.php
```diff
@@ -188,15 +188,20 @@ function register_term_columns() {
...
 			return;
 			
 		if ( !wp_verify_nonce( $_POST['nonce'], 'change-date' ) )
 			wp_die( $this->module->messages['nonce-failed'] );
 		
 		$current_user = wp_get_current_user();
 		$user_filters = $this->get_user_meta( $current_user->ID, self::usermeta_key_prefix . 'filters', true );
-		$user_filters['start_date'] = date( 'Y-m-d', strtotime( $_POST['ef-story-budget-start-date'] ) );
+		if ( isset( $_POST['ef-story-budget-today-submit'] ) ) {
+			$user_filters['start_date'] = date( 'Y-m-d' );
```
cojennin: What about using `current_time( 'Y-m-d' )` here? `date` wouldn't be accurate if the site's timezone was not UTC

cojennin: Or `date_i18n`

mjangda: 👍  for `current_time`

modules/story-budget/story-budget.php
```diff
@@ -188,15 +188,20 @@ function register_term_columns() {
...
+		} else {
+			$user_filters['start_date'] = date( 'Y-m-d', strtotime( $_POST['ef-story-budget-start-date'] ) );
```
mjangda: For consistent behavior:

```suggestion
			$user_filters['start_date'] = date_i18n( 'Y-m-d', strtotime( $_POST['ef-story-budget-start-date'] ) );
```

modules/story-budget/story-budget.php
```diff
@@ -322,11 +327,14 @@ function story_budget() {
...
 		
 		$output = '<form method="POST" action="' . menu_page_url( $this->module->slug, false ) . '">';
-			
+
 		$start_date_value = '<input type="text" id="ef-story-budget-start-date" name="ef-story-budget-start-date"'
 			. ' size="10" class="date-pick" value="'
-			. esc_attr( date_i18n( get_option( 'date_format' ), strtotime( $this->user_filters['start_date'] ) ) ) . '" /><span class="form-value">';
-		
+			. esc_attr( date_i18n( get_option( 'date_format' ), strtotime( $this->user_filters[ 'start_date' ] ) ) ) . '" />'
+			. '&nbsp;<input id="ef-story-budget-today-submit" name="ef-story-budget-today-submit" type="submit"'
```
mjangda: This should probably be a link rather than a button?

modules/story-budget/story-budget.php
cojennin: Ah interesting, this will need a patch similar to #546. `$_POST['ef-story-budget-start-date']` is not going to be parsed correctly by `strtotime` since it's value is `Y-m-d`

Either way, `date_i18n` seems more consistent, I can create a new PR for the localization issue

modules/story-budget/story-budget.php
cojennin: Yea, I'll make it a secondary button for the time being

PR #519, opened 2019-08-22
Post Author notifications UI fixes

### LINKED ISSUE #508: Unsubscribing author from Notifications is impossible: Show error or disable unsubscribe checkbox
### Expected behavior:

- When unticking the post author in `Notifications` metabox, they get unsubscribed
- OR If they aren't actually unsubscribed, we get a red flash instead of green after AJAX returns, and their box is still ticked.

### Current behavior:

- Unticking the post author causes a green flash, and the user seems to be unticked
- Reloading the page shows they have been re-ticked, it is impossible to unsubscribe the post's author

### Thoughts:

- This is really confusing to users, and likely to make them misunderstand the situation if they don't go back and look at the `Notifications` metabox after reloading
- I understand the reasons why the post author would remain always subscribed, and I don't think we need to change that aspect of this, it's understandable and users can work around it.
- BUT: if it's impossible to unsubscribe the author, we should make it clear when someone tries to untick them that it didn't work. 

### Proposed UX: AJAX returns red flash and leaves the user ticked

- Unticking the current author causes a red flash, and when AJAX returns, the user is still ticked. 
- This seems the easiest to do by editing how the AJAX response is handled.
- Downside is that if the author select menu on screen has been changed, but not saved, it is the "old" author that will be un-unsubscribeable, rather than the "new" one, since our AJAX can only see the saved author on the backend.

### Alternate UX: Disable the checkbox for current author based on a CSS class

- When rendering the user list, add a CSS class to the current post author: `post_following_list-author`
- CSS class matches the ones used to filer the notified list display,`like post_following_list-current_user` 
- Use this to disable the checkbox for the author so they can't be unticked. 
- Downside is the same as the AJAX solution: Doesn't account for the author having been changed during this pageload. 
-  Upside is that this seems really easy to implement. Adding a `disabled` property to the input that just lasts until the page is reloaded wouldn't involve much code at all.

### Alternate UX 2: Disable checkbox based on currently selected author on screen

- Same as previous UX, but we also have JS that updates the "disabled" user in Notifications based on who is currently selected as the author in the author metabox
- This seems a bit harder because now we need to update the checkbox whenever the author UI is changed, and involves external (changing) JS, unlike the first two solutions which only rely on `get_the_author()`-type unchanging API calls.
- Upside is that this will reflect the state AFTER saving better, because if you change the author AND unsubscribe the old one, then the old one will indeed remain unsubscribed
- FWIW I think that whole scenario of changing the author and unsubscribing the old one is VERY niche, and won't come up very often. Fixing the core problem of making it clear you can't unsubscribe the current author is a much higher priority.

### Can you replicate the key problem?

Maybe I'm missing something. Would be great just to hear if I'm right about the key premise, that it's impossible to unsubscribe the current author, and the UI lies and says that it is possible.

mikeyarce:
This PR will disable the notifications checkbox for `post_authors` if:

- The `ef_notification_auto_subscribe_post_author` filter is true (default behaviour)
- If the `post_author` is currently subscribed (there is an edge case where they might not be)

We also add a visual indication that the user is the Post Author:
Screen Shot 2019-08-22 at 2 33 50 PM

**Thoughts on implementation:**
My first attempt was to change how the user list is populated in `class-module.php`, but I quickly realized that looping through each user to programmatically add details is resource-intensive when you have many users.  I opted for JavaScript because this way we don't have to loop through each user, we can just target one specific user and modify the elements that way.

Fixes #508

### DIFF
--- a/modules/notifications/lib/notifications.css
+++ b/modules/notifications/lib/notifications.css
@@ -108,3 +108,11 @@
 #ef-usergroup-users h4 {
 	margin-top: 0;
 }
+
+.ef-post_following_list .post-author {
+	margin-right: 10px;
+	border: solid 2px #b8b8b8;
+	color: #b8b8b8;
+	padding: 4px;
+}
+
--- a/modules/notifications/lib/notifications.js
+++ b/modules/notifications/lib/notifications.js
@@ -31,25 +31,25 @@ jQuery(document).ready(function($) {
 	}
 
 	var maybe_disable_post_author_checkbox = function( ) {
-		if ( typeof post_author_is_follower == 'undefined' ) {
+		if ( typeof ef_post_author_is_follower == 'undefined' ) {
 			return;
 		}
-		if ( true == post_author_is_follower && true == post_author_auto_subscribe ) {
-			$('#ef-selected-users-' + post_author_id ).prop('disabled', true);
+		if ( true == ef_post_author_is_follower && true == ef_post_author_auto_subscribe ) {
+			$('#ef-selected-users-' + ef_post_author_id ).prop('disabled', true);
 		}
 	}
 
 	var display_post_author_warning = function( ) {
-		if ( typeof post_author_is_follower == 'undefined' ) {
+		if ( typeof ef_post_author_is_follower == 'undefined' ) {
 			return;
 		}
-		if ( true == post_author_is_follower && true == post_author_auto_subscribe ) {
-			$("label[for='ef-selected-users-" + post_author_id + "'] .ef-user-subscribe-actions").prepend( "<span class='post-author'>" + ef_notifications_localization.post_author + "</span>" );
+		if ( true == ef_post_author_is_follower && true == ef_post_author_auto_subscribe ) {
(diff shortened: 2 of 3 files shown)


dchymko:
Tested successfully in conjunction with #540.  It makes sense to merge these together, as that will allow users to unsubscribe from their own posts which I don't think we want.

cojennin:
Looking good. Couple thoughts:

1. When using the Block Editor, the disabled checkbox won't appear after the post is saved until after the page is reloaded. Can we observe a post save and trigger the disabled checkbox?
2. It can also be the case that the current user will be auto-subscribed to the post if the filter `ef_notification_auto_subscribe_current_user` returns `true`. Can we add a similar UI treatment for the current user (i.e., show checkbox as disabled when post is saved and add a nice box with something like "Auto-Subscribed")?
3. I really like this UI treatment, does anyone have any reasons as to why we can't just show it all the time instead of under certain conditions? It adds nice context to the notifications list and it'd be easier to maintain without logical checks

mjangda:
Closing; continued in #563

PR #520, opened 2019-08-22
Remove auto-subscribe of current users to post notifications

mikeyarce:
This change means the current user will no longer be auto-subscribed to post notifications when actions are taken on a post.


dchymko:
> _That said_ there's a separate bug that makes this situation worse: It seems to be impossible to untag yourself from a post. When I uncheck my box after being auto-subscribed it flashes green and the box returns unchecked, but reloading the page shows me checked again. Not a good situation at all. Clearly a bug.

This bug has been addressed in #540. As far as the bigger issue around adding the ability to customize notifications, perhaps we can create that as an enhancement issue.

PR #521, opened 2019-08-23
Fix setting initial status in Gutenberg

### LINKED ISSUE #514: "Pitch" status turns to "Draft" on save
Steps to reproduce:

1) Create new post
2) Hit save with "Pitch" selected
3) Notice that "Draft" is the post status now instead of "Pitch".

mikeyarce:
When you open a new post, click “Save Draft” - it will save the post as a Draft even if a custom status is selected.

This fixes that by setting the status right away instead of waiting for a status to be selected from the dropdown.

edit-flow-status

There is one minor outstanding issue:  when the post is done saving, the text goes back to "Save Draft", but it does actually save the right status. I'll add some ideas on what to do in another comment.

Fixes #514


mjangda:
> I'll add some ideas on what to do in another comment.

@mikeyarce do you still remember these ideas? :)

mjangda:
Tested and I'm finding that for existing posts, this reverts the custom status (to the first in the list) when you refresh the page (i.e. if you set the post to "In Progress", then save, then refresh, the dropdown and status will be set to "Pitch" which is top of the status list).

blocks/src/custom-status/block.js
```diff
@@ -24,6 +24,15 @@ let sideEffectL10nManipulation = status => {
...
 }
 
+// Set the status to the default custom status.
+subscribe(function () {
+  var status = select('core/editor').getEditedPostAttribute('status');
+  if ( typeof status !== 'undefined' && status !== 'publish' ) { 
+    sideEffectL10nManipulation( getStatusLabel( status ) );
+  }
+})
+dispatch('core/editor').editPost( { status: ef_default_custom_status });
```
mjangda: This line is breaking the status revert described here: https://github.com/Automattic/Edit-Flow/pull/521#issuecomment-548836488

mjangda:
Okay, should be fixed now.

We're no longer dispatching on load. For existing posts, this ends up re-writing their status. Instead, we now limit the updates to new posts.

Also removes the status from "save" button, which creates for awkward experience where the grammar doesn't quite work (e.g. "Save Assigned" or "Save Waiting for Feedback"). This also minimizes the "Save Draft" bug mentioned in the PR.

blocks/src/custom-status/block.js
mjangda: Fixed.

PR #532, opened 2019-10-30
Update PHP 7 version in TravicCI build Matrix

dchymko:
Updates the CI build versions to use current and previous version of PHP 7.x

Note: `nightly` cannot currently work due to a conflict between PHPUnit dependencies and the WordPress version requirements.  Nightly (PHP 8.0.0-dev) requires PHPUnit 8+ but the WordPress test suite only works with PHPUnit 5.7 or older.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,8 @@ language: php
 
 php:
     - "5.6"
-    - "7.0"
-    - "7.1"
+    - "7.2"
+    - "7.3"
 
 services:
     - mysql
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -2,44 +2,44 @@ language: php
...
       env: WP_VERSION=5.2 WP_MULTISITE=0
     - php: "5.6"
       env: WP_VERSION=5.2 WP_MULTISITE=1
-    - php: "7.0"
+    - php: "7.2"
       env: WP_VERSION=5.2 WP_MULTISITE=1
-    - php: "7.1"
+    - php: "7.3"
       env: WP_VERSION=5.2 WP_MULTISITE=1
     - php: "nightly"
```
mjangda: I wonder if we should just remove php nightly? Not really that much benefit to testing it.

dchymko: Agree, if anything we should add PHP 7.4

PR #538, opened 2019-10-31
Wrap network check with block to ensure function exists

dchymko:
Bring in the changes from #490

While testing this, I encountered compatibility issues between Ninja Shop and the Custom Types module in Edit Flow. Logging that as a separate issue

### DIFF
--- a/modules/user-groups/user-groups.php
+++ b/modules/user-groups/user-groups.php
@@ -690,7 +690,8 @@ class EF_User_Groups extends EF_Module {
 		// `get_current_screen()` is defined on most admin pages, but not all.
 		if( function_exists( 'get_current_screen' ) ){
 			//Don't allow update of user groups from network
-			if ( ( !is_null( get_current_screen() ) ) && ( get_current_screen()->is_network ) ) {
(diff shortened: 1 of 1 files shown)


modules/user-groups/user-groups.php
```diff
@@ -687,9 +687,13 @@ function user_profile_update( $errors, $update, $user ) {
 		if ( !$update )
 			return array( &$errors, $update, &$user );
 
-		//Don't allow update of user groups from network
-		if ( ( !is_null( get_current_screen() ) ) && ( get_current_screen()->is_network ) )
-			return;
+		// `get_current_screen()` is defined on most admin pages, but not all.
+		if( function_exists( 'get_current_screen' ) ){
+			//Don't allow update of user groups from network
+			if ( ( !is_null( get_current_screen() ) ) && ( get_current_screen()->is_network ) ) {
```
mjangda: Minor suggestion to fix the double function call and improve readability a bit:

```suggestion
			$screen = get_current_screen();
			if ( ! is_null( $screen ) && $screen->is_network ) {
```

PR #546, opened 2019-11-04
Fix/date picker localization

### LINKED ISSUE #457: Cannot save Date meta field with international name
If a website in danish use Edit-Flow and the Datepicker picks up on this and returns the danish version of "May" (Maj), then the date will not be saved on post update.
Most other months work fine since it's the same shorthand of the months name. I'm figuring this is a scenario with other languages as well.

Would the solution be to force the datepicker to be english, as a quick solution?

cojennin:
## Description

Fixes #457. Instead of trying to parse the localized date (which fails when using `strtotime` or `DateTime::createFromFormat` for certain cases), we're setting the DateTime numerically so we can parse it.

Also, we're adding a z-index to the DatePicker, it was hidden when using the BlockEditor

## Steps to Test

1. In `Settings`, set language to `Dansk`
2. Select a post, and modify a `date` editorial metadata and select a date in May
3. Save the post, and confirm that the date is saved

Previously, a date in the month of May would not save (because `strotime` and `DateTime:createFromFormat` would fail to parse it) when the language was `Dansk`. This was probably impacting other languages as well


cojennin:
Part of this PR also encompasses fixing up the PHPUnit testing to enforce better ordering (AJAX unit tests define `DOING_AJAX` and need to happen last to prevent interference with other test), see [wordpress-develop](https://github.com/WordPress/wordpress-develop/blob/master/phpunit.xml.dist#L13) for reference

phpunit.xml.dist
```diff
@@ -0,0 +1,26 @@
...
+	convertNoticesToExceptions="true"
+	convertWarningsToExceptions="true"
+	>
+	<testsuites>
+		<testsuite name="default">
+			<directory suffix=".php">./tests/</directory>
+			<exclude>tests/test-edit-flow-custom-status-ajax.php</exclude>
+		</testsuite>
+		<!-- Sets the DOING_AUTOSAVE constant, so needs to be run last -->
+		<testsuite name="autosave">
```
mjangda: Rather than relying on order which is pretty fragile, it's better to use https://phpunit.de/manual/6.5/en/appendixes.annotations.html#appendixes.annotations.runTestsInSeparateProcesses to avoid leaking state between suites.

cojennin: Makes sense, I was just stealing from [wordpress-dev](https://github.com/WordPress/wordpress-develop/blob/master/phpunit.xml.dist#L13).

It's the `require_once` that's setting the `DOING_AJAX`, moving it into [setUp](https://github.com/Automattic/Edit-Flow/pull/546/commits/365511b5c30c83dd907cc454e15b570f0640874c#diff-df02cad6e7abc4e31f5cab9e94323a24R11) and adding the `runTestsInSeparateProcesses` seems to work

cojennin:
@mjangda made one [small change](https://github.com/Automattic/Edit-Flow/pull/546/commits/ffded67d23fb5d5801f9af050479dfb77bc743c5) that was causing test [failures on PHP 5.6](https://travis-ci.org/Automattic/Edit-Flow/jobs/626502808?utm_medium=notification&utm_source=github_status).

PR #547, opened 2019-11-05
Update publishing instructions

mjangda:
Misc improvements to the docs and release script after pushing a release.

### DIFF
--- a/.svnignore
+++ b/.svnignore
@@ -1,3 +1,4 @@
+*.map
 BLOCKS.md
 CONTRIBUTING.md
 composer.json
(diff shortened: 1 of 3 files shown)


.svnignore
```diff
@@ -7,8 +8,6 @@ package.json
 phpunit.xml
 PUBLISHING.md
 README.md
-vipgo-helper.php
-wpcom-helper.php
```
cojennin: Why remove these two?

mjangda: Yeah, I realized that mistake after and restored them in a future PR.

cojennin: Cool, going to merge

mjangda: Oh, sorry, I totally misunderstood the question. We want to include these files for any VIPs that download the plugin from the .org repo (rather than GitHub).

Issue #505, opened 2019-04-09
Calendar: Object cache not cleared


asadowski10:
Hi team,

Since the 0.8.3 release and the feature > Improvement: Cache calendar items for each user individually to prevent potential cache pollution. Props justnorris.
We are having cache problems.
When we update content, the cached version is not updated.

The cache key is based on the user_id ( $cache_key = $post->ID . $can_modify . '_' . get_current_user_id(); ) but when you delete the cache , the key has no user_id  ($post_id . 'can_modify' )
https://github.com/Automattic/Edit-Flow/blob/master/modules/calendar/calendar.php#L1800

Could you delete cache for user or remove user_id in cache key to have cache for all users ?

Thanks a lot

mjangda:
The change was introduced here #412 to make the cache user-specific. Removing the user ID would re-introduce the bug it was trying to fix.

Two thoughts:
- Find a way to clear the cache for all users (otherwise we risk showing stale data to some).
- See if we can safely remove the HTML fragment caching and avoid all this complexity.

jerclarke:
Wow this is a real nightmare to have gone on for so long. Found this ticket after persistent complaints from our sub-editors about the calendar being wrong in very random ways, and now it all makes sense. 

To clarify the form this takes for anyone not familiar: The calendar item for a specific post will get trapped with a certain version of the metadata for the post (title+status) even after the post is edited. Notably the date of the post is **not** effected. If you drag a post to a new date, or change the date in the post editor, then reload the calendar, the item will show on the correct calendar day. BUT if you change the status or the title, it will show on the new date with the old title/status. 

After setting up two windows with different users to confirm this, I commented out the lines that return the cached value in `generate_post_li_html()`, reloaded the calendar, and the data was all immediately correct. 

Very chaotic! No wonder I had trouble pinning down exactly what was wrong!

Steps to reproduce:
* Log in as a user that can edit all posts
* In a private browser, log in as another user that can edit all posts
* Open the calendar in both windows (I believe it will be accurate at this point)
* Edit a post to have a different title and status in one of the windows
* Reload both windows: They both have the old title and status for that post. 

I'm going to try to figure out a solution to purge the cache for everyone, or else I might just disable the cache entirely, as it seems like it might be more effort than it's worth. 

**Anyone know what about this `generate_post_li_html()` is so desperately in need of caching that it's worthwhile to cache nonces?**

[...]

PR #548, opened 2019-11-12
Rework caching, no longer cache html but cache expensive queries

cojennin:
First pass at improving #505 

Instead of caching the entirety of the html, cache some of the queries that go into building the post details screen (and aren't user dependent). I added about 300 posts to the calendar to check, and there are some small improvements. Have not seen any major performance degradation, though [this function](https://github.com/Automattic/Edit-Flow/blob/fix/calendar-caching/modules/notifications/notifications.php#L289) now goes uncached.

I'd prefer not to remove the cache in it's entirety without some further investigation, but this PR starts to untangle some of it from the calendar. I've also tried to improve some of the calls being made to other functions to eke out some small performance improvements, and consolidate some calls to functions that didn't seem necessary

Eventually, we could consider doing some things in future PR's like:
1. Limiting the posts that get rendered per day
2. Defer fetching details of a post until a user clicks the post
3. Rendering the initial post information (i.e., title and type) client side

### DIFF
--- a/modules/calendar/calendar.php
+++ b/modules/calendar/calendar.php
@@ -21,7 +21,7 @@ class EF_Calendar extends EF_Module {
 	var $max_visible_posts_per_date = 4; // total number of posts to be shown per square before 'more' link
 
 	private $post_date_cache = array();
-	private static $post_li_html_cache_key = 'ef_calendar_post_li_html';
+	private static $post_li_details_cache_key = 'ef_calendar_post_li_html';
 
 	/**
 	 * Construct the EF_Calendar class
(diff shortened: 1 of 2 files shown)


mjangda:
As discussed, this needs a bit more testing and profiling to see if this is a beneficial change (quick testing shows some benefits).

jerclarke:
Really too bad this died on the vine, considering the bug it tries to fix is absolutely devastating to the trustworthiness of the calendar, and remains to this day. 

Long shot but @cojennin, do you remember how bad the actual hit was of turning off caching v. leaving it on? Does it break the calendar? Because the bug currently caused by the per-user cache breaks the calendar.

GaryJones:
Closing this PR as the underlying issue has been resolved differently.

## Background

This PR attempted to fix #505 by reworking the caching strategy - caching expensive queries instead of the full HTML output. While this was a reasonable approach, further investigation (documented in #756) revealed that:

1. The caching wasn't providing meaningful performance benefits
2. The cache invalidation was fundamentally broken (keys didn't match between set and delete operations)
3. The real performance issue was the `wp_dropdown_categories()` calls for hierarchical taxonomies, not database queries

## Resolution

- **#828** removed the broken `li_html_cache` entirely, solving the stale cache problem reported in #505
- **#757** made hierarchical taxonomies non-editable, solving the performance issue caused by loading full category trees for every post
- **#829** proposes a future enhancement with AJAX-loaded modals that could implement proper caching if needed

The ideas in this PR (like deferring post details until click) are captured in #829 for future consideration.

Thank you @cojennin for the original investigation and approach - it helped inform the eventual solution!

Issue #551, opened 2019-11-14
WP menu's unusable on v5.3 using plugin


fivestep:
Hello,
I found this 'notice' with the latest WP version '5.3'. This is found in the menu's page, which makes adding items to the menu impossible. 

> ( ! )  Notice: Trying to get property 'ID' of non-object in  /var/www/html/wp-content/plugins/edit-flow/modules/custom-status/custom-status.php  on line 756

**Custom_status.php**
<img width="580" alt="image" src="https://user-images.githubusercontent.com/6503303/68899370-1eb3a580-06f7-11ea-95f7-c8496de94ff0.png">
Inside function **check_if_post_state_is_status()** I ran **var_dump($post->ID);** which returns with null, but as soon as I downgrade WP to version **5.2.4**, the **var_dump** is no longer being called there in the menus page.

<img width="370" alt="image" src="https://user-images.githubusercontent.com/6503303/68898977-41918a00-06f6-11ea-805a-619c778385dd.png">

<img width="837" alt="image" src="https://user-images.githubusercontent.com/6503303/68898933-20c93480-06f6-11ea-925b-d2d67655351b.png">

cojennin:
This should be resolved in the newest version of Edit Flow (0.9.3). Please update when you have a moment and let us know if this problem persist. Closing this for now, thanks!

PR #552, opened 2019-11-15
Fix WP Menu post title notice

cojennin:
Fix for #551 

`display_post_states` has two arguments, one of them `$post` (since WP 3.6). Use the argument instead of the global `$post`


mjangda:
As discussed: We should consider splitting this up into two separate PRs (the display_states_fix and em dash).

For the em dash, we should look to see if we can e2e tests.

Issue #553, opened 2019-11-18
Missing item in EditFlowCustomStatuses prevents scheduling posts


nickyout:
On the post.php of a specific post with &action=edit I see a lot of the same uncaught rejected promises. Cannot read property 'label' of undefined. It prevents me from scheduling posts.

## Expected/Desired Behavior

Ignoring the error, I would like to be able to schedule my posts as usual. 

## Actual Behavior

The schedule button throws the abovementioned error in the console. From the user's standpoint clicking the button does nothing.

## Steps to Reproduce the Problem

1. Open devtools (F12)
1. Edit a post
1. Click Publish...
1. Select a future date
1. Click Schedule or whatever the english translation is. Inplannen in nl. 
1. Go to step 5. 

## (Optional) Additional notes

Attempting to further understand the problem, I have found out the following:

- It is version 0.9.1
- The direct reason for the error is that in blocks/src/custom-status/block.js it is expected that `window.EditFlowCustomStatuses` has an item for slug 'future', but it does not.
- I guess this js var is set by modules/custom-status/compat/block-editor.php, which means that
- Probably modules/custom-status/custom-status.php is the source of the value, which can be appended by handle_add_custom_status, but
- Using name 'future' is restricted and will not work from the EditFlow configuration UI. And at this point, I got scared that adding 'future' with brute force would break things.

Full disclosure: I am not the owner of the (website having the) problem. I am not a wp/php programmer, I just read/looked up what it probably does.

I patched the problem by simply adding `|| {label: '???'}` after the `find` in the javascript file. 

My first naive suggestion to a fix would be to replace `window.EditFlowCustomStatuses` in the js file with `window.custom_statuses` as it seems to have the missing slug 'future' with the appropriate translation for 'label' and the other 5 custom statuses that were in `EditFlowCustomStatuses`.

If I am missing something in some kind of configuration, please let me know.

mjangda:
Thanks for such a detailed report, @nickyout (and sorry for the trouble). We'll get this patched and released ASAP.

davisshaver:
+1 to getting this out quickly. Here is patch that @nickyout mentioned although as he suggest probably a better way to do this. https://github.com/Automattic/Edit-Flow/pull/554

mjangda:
A fix for this have been pushed out with 0.9.2. Thanks for all your help with this @nickyout @davisshaver @cojennin!

We're going to continue a better, long-term fix in #555.

PR #554, opened 2019-11-20
553: Add patch to re-enable scheduling in block ed

davisshaver:
Temporary patch as described in https://github.com/Automattic/Edit-Flow/issues/553.


davisshaver:
Carried forward in https://github.com/Automattic/Edit-Flow/pull/555.

PR #555, opened 2019-11-20
Fix custom status label bug

cojennin:
We should be using the status in the `Save as` section of the Gutenberg editor. This change will add the status to the `Save as` section when a user saves the post with a custom status. There's currently no way to hook into core to do this, so we have a timeout to correct the status once it's set by core.

### To validate status being set
- Create a post
- Change it's status to "Assigned"
- Click "Save as Assigned"
- Confirm that after saving, "Save as Assigned" still appears


mjangda:
Note that we need to test these changes with the Classic Editor as well.

mjangda:
Aside: we should revert the changes to package-lock.json, since those aren't necessary for this PR.

davisshaver:
Tested on site with block editor enabled for posts and can confirm this branch addresses issue described in #553. Also confirmed the `Move to Trash` error described above exists on `master` and is addressed by the same branch.. 

I also tested with post type that does not have block editor enabled but does have custom statuses enabled. Can confirm that bugs with scheduling and trashing posts does _not_ seem to be present on `master`. Additionally, no regression with scheduling and trashing experienced was experienced with this classic editor post type when testing with `fix/custom-status-scheduled`.

modules/custom-status/custom-status.php
```diff
@@ -416,32 +416,14 @@ function post_admin_header() {
...
-				'description' => '',
-			);
-
-			// Load the custom statuses
-			foreach( $custom_statuses as $status ) {
-				$all_statuses[] = array(
-					'name' => esc_js( $status->name ),
-					'slug' => esc_js( $status->slug ),
-					'description' => esc_js( $status->description ),
+			$all_statuses = array_merge( $custom_statuses, $this->get_default_statuses() );
```
davisshaver: +1 to abstracting this into `get_default_statuses()`

modules/custom-status/custom-status.php
```diff
@@ -650,6 +632,42 @@ function get_custom_statuses( $args = array() ) {
 		return $ordered_statuses;
 	}
 
+	/**
+	 * The default statuses from core. Returned as an array of stdClass to 
+	 * make it easy to merge with our custom statuses (which are WPTerm objects)
+	 * 
+	 * return array an array of stdClass objects
```
davisshaver: missing `@` in return

blocks/src/custom-status/block.js
```diff
@@ -11,19 +11,37 @@ let { SelectControl } = wp.components;
...
 // @see https://github.com/WordPress/gutenberg/issues/3144
-let sideEffectL10nManipulation = status => {
-  let node = document.querySelector('.editor-post-save-draft');
-  if ( node ) {
-    document.querySelector( '.editor-post-save-draft' ).innerText = `${ __( 'Save' ) }`
-  }
+// Gutenberg will also override the status set in '.editor-post-save-draft' after save, and there isn't yet a way
+// to subscribe to a "post save" message. So instead set a timeout and override the text in '.editor-post-save-draft'
+let schedulePostStatusUpdater = () => {
+  setTimeout(() => {
```
davisshaver: Potentially a more "native" way of doing this would be to monitor for changes to the modified time (set modified time on load and compare/update/fire action in `wp.data.subscribe` callback.

modules/custom-status/custom-status.php
```diff
@@ -650,6 +632,42 @@ function get_custom_statuses( $args = array() ) {
...
+		$private_status = new stdClass;
+		$private_status->name = __( 'Privately Published', 'edit-flow' );
+		$private_status->slug = 'private';
+		$private_status->description = '';
+
+		$scheduled_status = new stdClass;
+		$scheduled_status->name = __( 'Scheduled', 'edit-flow' );
+		$scheduled_status->slug = 'future';
+		$scheduled_status->description = '';
+
```
davisshaver: minor but might be an extra linebreak here?

modules/custom-status/custom-status.php
```diff
@@ -650,6 +632,42 @@ function get_custom_statuses( $args = array() ) {
...
+
+		return [ $published_status, $private_status, $scheduled_status ];
+	}
+
+	/***
```
davisshaver: Minor but maybe an extra asterisk here?

blocks/src/custom-status/block.js
```diff
@@ -11,19 +11,37 @@ let { SelectControl } = wp.components;
...
+    let node = document.querySelector('.editor-post-save-draft');
+    if ( node ) {
+      document.querySelector( '.editor-post-save-draft' ).innerText = `${ __( 'Save as' ) } ${statusLabel}`
+    }
+
+    schedulePostStatusUpdater();
+  }, 250);
 }
 
+schedulePostStatusUpdater();
```
mjangda: I'm a bit worried out the unwanted (and unexpected) effects of having an infinite loop to fix the label bug (i.e. where the save button occasionally flips to "Save Draft" after saving).

It would be better to keep the change to have the save button as generic as possible (i.e. "Save") and live with the occasional text flip. And then work with the Gutenberg project to find a better way to fix this.

mjangda:
@cojennin I have a much simpler fix in #556 that I'm going to push out. I'm going to keep this open as I think it's worth iterating on to improve the the "Save" button behavior as it's definitely not ideal right now (@davisshaver's idea is worth exploring, for example).

blocks/src/custom-status/block.js
cojennin: Yea, not a bad thought, I'll mess around and see if I can get that working

blocks/src/custom-status/block.js
cojennin: Yea, maybe there's a better way to approach this while waiting for Gutenberg, I'll mess around a bit. Just seems like a point of confusion for consumers

modules/custom-status/custom-status.php
cojennin: Still kinda confusing, but hoping to move in the right direction

modules/custom-status/custom-status.php
```diff
@@ -452,7 +434,7 @@ function post_admin_header() {
 			// Now, let's print the JS vars
 			?>
 			<script type="text/javascript">
-				var custom_statuses = <?php echo json_encode( $all_statuses ); ?>;
+				var custom_statuses =  JSON.parse( decodeURIComponent( '<?php echo rawurlencode( wp_json_encode( $statuses_js ) );  ?>' ) );
```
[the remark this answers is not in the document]
cojennin: Same, but it's in the [suggested VIP code sample for encoding JSON](https://github.com/Automattic/vip-code-samples/blob/adeee777b3b3df6c5ec2193d5d9460149b9211e2/10-security/js-dynamic.php#L12). Down to move it back to just `wp_json_encode` if no concerns there

modules/custom-status/custom-status.php
cojennin: Ah, good catch

blocks/src/custom-status/block.js
cojennin: Though I'm not sure it'll solve the underlying issue. We can observe [post update](https://github.com/Automattic/Edit-Flow/blob/master/blocks/src/custom-status/block.js#L89), but the problem is [here](https://github.com/WordPress/gutenberg/blob/95e769df1f82f6b0ef587d81af65dd2f48cd1c38/packages/editor/src/components/post-saved-state/index.js#L38-L41).

Gutenberg is setting a timeout of 1s to flip the message from "Saving..." to the [original text](https://github.com/WordPress/gutenberg/blob/95e769df1f82f6b0ef587d81af65dd2f48cd1c38/packages/editor/src/components/post-saved-state/index.js#L104): "Save as `Draft|Pending`". But if there's concern over about polling, we could set one timeout for something like 1.2s and that might solve it (but small race condition)

davisshaver: I get what you mean. The two options you proposed (callback loop and timeout after save) make sense to me based on current Gutenberg code. This seems like it could also be a good place for an upstream change to Gutenberg if there's a filter that could make it work. Best idea I have right now would be providing a way to filter/bring your own function to use instead of `isPending ? __( 'Save as Pending' ) : __( 'Save Draft' );`

cojennin:
Yea, an upstream change to Gutneberg would definitely be preferred (less
work for us!).

In the meantime, there might be a better way than this infinite polling
approach, it’s worth hacking around on to see what we can do

cojennin:
@mjangda that fix looked good! It was what I had originally started with before trying to see if we could resolve some of these other issues

cojennin:
@mjangda @davisshaver doesn't seem like there's a real clean way to overcome Gutenberg explicitly setting `Save as Draft` or `Save as Pending` [after a user clicks save](https://github.com/WordPress/gutenberg/blob/95e769df1f82f6b0ef587d81af65dd2f48cd1c38/packages/editor/src/components/post-saved-state/index.js#L104) after a [1s timeout](https://github.com/WordPress/gutenberg/blob/95e769df1f82f6b0ef587d81af65dd2f48cd1c38/packages/editor/src/components/post-saved-state/index.js#L39). 

I was thinking we might be able to use a [MutationObserver](https://caniuse.com/#search=MutationObserver) instead of polling, but browser support isn't great (see IE), and the [shim I found](https://www.npmjs.com/package/mutationobserver-shim) effectively does exactly what this PR does now (set's a timeout to observer a DOM change).

I added an `attempts` parameter to the timeout function to prevent looping indefinitely.

It would be great to get a fix into Gutenberg core for this issue, but in the meantime it'd be great to solve this for Edit Flow users. [The behaviour we have now](https://imgur.com/a/8IxSPIp), where the text flips from "Save" to "Save Draft" after a user clicks save, regardless of status, is confusing.

davisshaver:
I dropped an issue into Gutenberg, linking here for visibility: https://github.com/WordPress/gutenberg/issues/18743

cojennin:
@mjangda @davisshaver I've updated this PR with some e2e testing to verify the PR works as expected, let me know how're you're feeling about it. I'm a little more confident now that it'll work as expected

cojennin:
Closing in favor of #585

PR #559, opened 2019-11-25
Adding phpcs as part of Travis build process

cojennin:
This change executes PHPCS (`WordPress-VIP-Go`) on Travis. It should execute only on changed lines of code. It will output the line and rule violation that occurred in the console of the Travis build. You can also run it from your local machine after committing a change (could run on staged diff as well, with changes).

Eventually, we may want to:

- Introduce a larger coverage scope (run PHPCS over a whole file, not just the changes)
- Make this part of a pre-commit hopok

But in the meantime, it should introduce a lightweight sanity check for new changes to ensure we're not introducing any new violations without acting as a blocker

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ services:
     - mysql
 
 env:
-    - WP_VERSION=latest WP_MULTISITE=0 #Current stable release
+    - RUN_PHPCS=1 WP_VERSION=latest WP_MULTISITE=0 #Current stable release
 
 # Test WP 5.1 and 5.2 and on PHP 5.6, 7.2 and 7.3 (w and w/o multisite enabled)
 
@@ -41,6 +41,12 @@ matrix:
     - php: "7.3"
       env: WP_VERSION=5.2 WP_MULTISITE=1
 
+install:
+    - |
+      if [[ ${TRAVIS_PHP_VERSION:0:2} == "7." ]]; then
+        composer install
+      fi
+
 before_script:
     - bash bin/install-wp-tests.sh wordpress_test root '' localhost $WP_VERSION
     - export COMPOSER_BIN_DIR="$HOME/.config/composer/vendor/bin"
@@ -52,6 +58,11 @@ before_script:
     - phpunit --version
     - php --version
 
-script: phpunit
+script: 
+  - |
+      if [[ ${TRAVIS_PHP_VERSION:0:2} == "7."  ]] && [[ ${ RUN_PHPCS } == "1" ]]; then
+        bash bin/phpcs-diff.sh
+      fi
+  - phpunit
 
 sudo: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,7 +60,7 @@ before_script:
 
 script: 
   - |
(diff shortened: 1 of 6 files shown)


mjangda:
Nice! Where does the `diffFilter` bin come from?

cojennin:
Note, this will only run on Travis environments with PHP 7 or greater (due to a transitive dependency requirement). The output will look something like:

Screen Shot 2019-11-25 at 4 01 14 PM

In this screenshot we get 100% coverage since there were 

- No PHP changes
- No issues with PHPCS

If there were PHPCS issues, you'd see something like the following (run from my local computer on a new commit, with an intentional error placed in the code):

Screen Shot 2019-11-25 at 4 12 19 PM

cojennin:
@mjangda the `diffFilter` bin comes from [this library](https://github.com/exussum12/coverageChecker). I was messing around with some alternatives (both called phpcs-dff: [1](https://github.com/zaantar/phpcs-diff) and [2](https://github.com/olivertappin/phpcs-diff) but the `coverageChecker` library seemed to work the best. Open to exploring some alternatives if you have any suggestions, was just googling around to find something that might work

.travis.yml
```diff
@@ -52,6 +58,11 @@ before_script:
     - phpunit --version
     - php --version
 
-script: phpunit
+script: 
+  - |
+      if [[ ${TRAVIS_PHP_VERSION:0:2} == "7." ]]; then
```
mjangda: I think we should put this behind an env var and only run on a single suite (no point running `n` times and slowing all of them down)

```suggestion
      if [[ ${TRAVIS_PHP_VERSION:0:2} == "7." && ${ RUN_PHPCS } == "1" ]]; then
```

And update line 11 to:

```
    - RUN_PHPCS=1 WP_VERSION=latest WP_MULTISITE=0 #Current stable release
```

bin/phpcs-diff.sh
```diff
@@ -0,0 +1,13 @@
...
+DIFF_FILE=$(mktemp)
+PHPCS_FILE=$(mktemp)
+
+git remote set-branches --add origin master
+git fetch
+git diff origin/master... > $DIFF_FILE
+
+$DIR/../vendor/bin/phpcs --standard=phpcs.xml.dist --report=json > $PHPCS_FILE || true 
+
+$DIR/../vendor/bin/diffFilter --phpcs $DIFF_FILE $PHPCS_FILE 0
```
mjangda: 419c4f says this:

> Settin[g] coverage to 0 so the Travis build doesn't fail on PHPCS issues

Shouldn't we get Travis to fail to force us and catch and fix the errors? Otherwise we'll just ignore them :)

Or is there some nuance about how diffFilter failures trigger?

cojennin: Ha, yea, not sure what kind of false positives it might kick up (haven't used diffFilter before), so was gonna hang off on making it a hard stop until we get a look at what it produces

mjangda: Cool, makes sense.

mjangda:
:shipit:

Excited to see how well this works (and get it in use across all our WP repos).

PR #560, opened 2019-11-26
Ensure status_friendly_names are provided for notifications

cojennin:
Some extra work on top of @jerclarke's work in #502 to add tests and account for cases where a `status_obj` would be null.


tests/test-edit-flow-notifications.php
```diff
@@ -0,0 +1,109 @@
+<?php
+
+class WP_Test_Edit_Flow_Notifications extends WP_UnitTestCase {
+
+	protected static $admin_user_id;
+	protected static $EF_Notifications;
```
mjangda: Minor: should probably be lower-case?

```suggestion
	protected static $ef_notifications;
```

And updated throughout the class.

tests/test-edit-flow-notifications.php
```diff
@@ -0,0 +1,109 @@
...
+		
+		self::$EF_Notifications = new EF_Notifications();
+		self::$EF_Notifications->install();
+		self::$EF_Notifications->init();
+	}
+
+	public static function wpTearDownAfterClass() {
+		self::delete_user( self::$admin_user_id );
+		self::$EF_Notifications = null;
+	}
```
mjangda: Any reason not to call the parent function as well?

```suggestion
		parent::wpTearDownAfterClass();
	}
```

tests/test-edit-flow-notifications.php
```diff
@@ -0,0 +1,109 @@
+<?php
+
+class WP_Test_Edit_Flow_Notifications extends WP_UnitTestCase {
+
+	protected static $admin_user_id;
+	protected static $EF_Notifications;
+		
+	public static function wpSetUpBeforeClass( $factory ) {
```
mjangda: Any reason not to call the parent function as well?

```suggestion
	public static function wpSetUpBeforeClass( $factory ) {
		parent::wpSetUpBeforeClass();
```

tests/test-edit-flow-notifications.php
```diff
@@ -0,0 +1,109 @@
...
+
+		global $pagenow;
+		$pagenow = 'post.php';
+	}
+
+	function tearDown() {
+		parent::tearDown();
+
+		global $pagenow;
+		$pagenow = 'index.php';
```
mjangda: Rather than assuming that the pagenow should be set to `index.php` we should track what the value was at `setUp` and restore it to that.

tests/test-edit-flow-notifications.php
```diff
@@ -0,0 +1,109 @@
...
+
+	function setUp() {
+		parent::setUp();
+
+		global $pagenow;
+		$pagenow = 'post.php';
+	}
+
+	function tearDown() {
+		parent::tearDown();
```
mjangda: As a best practice, I think it's best to call `parent::tearDown()` after our logic has run (as wp-tests does other resets that we may inadvertently impact).

tests/test-edit-flow-notifications.php
```diff
@@ -0,0 +1,109 @@
...
+		 */
+		$typenow = 'post';
+
+		$edit_flow->custom_status->module->options->post_types = array( 'page' );
+		/**
+		 * Initiate a full cycle (install/init) to ensure the core statuses are returned
+		 * instead of custom stautses (since we're disabling the module for this post type)
+		 */
+		$edit_flow->custom_status->install();
+		$edit_flow->custom_status->init();
```
mjangda: Oh, Edit Flow; you don't make things easy :)

tests/test-edit-flow-notifications.php
```diff
@@ -0,0 +1,109 @@
...
+		 * since the tests don't trigger `admin_init` and the `install` function is where
+		 * the custom post type taxonomy is loaded into the DB. 
+		 * 
+		 * So make sure we do one cycle of `install` followed by `init` to ensure 
+		 * custom post type taxonomy has been loaded.
+		 */
+		$edit_flow->custom_status->install();
+		$edit_flow->custom_status->init();
+
+		self::$admin_user_id = $factory->user->create( array( 'role' => 'administrator' ) );
```
mjangda: In general, probably better to create the user on `setUp` so it's reset between tests?

modules/notifications/notifications.php
```diff
@@ -586,6 +586,28 @@ function notification_status_change( $new_status, $old_status, $post ) {
...
+			/**
+			 * get_post_status_object will return null for certain statuses (i.e., 'new')
+			 * The mega if/else block below should catch all cases, but just in case, we 
+			 * make sure to at least set $old_status_friendly_name and $new_status_friendly_name
+			 * to an empty string to ensure they're at least set.
+			 * 
+			 * Then, we attempt to set them to a sensible default before we start the
+			 * mega if/else block
+			 */
+			if ( !is_null( $old_status_post_obj ) ) {
```
mjangda:
```suggestion
			if ( ! is_null( $old_status_post_obj ) ) {
```

modules/notifications/notifications.php
```diff
@@ -586,6 +586,28 @@ function notification_status_change( $new_status, $old_status, $post ) {
...
+				$old_status_friendly_name = $old_status_post_obj->label;
+			}
+			
+			if ( !is_null( $new_status_post_obj ) ) {
```
mjangda:
```suggestion
			if ( ! is_null( $new_status_post_obj ) ) {
```

tests/test-edit-flow-notifications.php
cojennin: From [poking around](https://github.com/Automattic/jetpack/blob/384678b75d1ccdbb3f8bdd23bff21c9d6045a4fb/tests/php/core-api/wpcom-fields/test-class.subscribers-endpoint.php#L15) it seemed like most folks were creating users in `wpSetUpBeforeClass` and in individual tests (would `$factory` be available in `setUp`?)

tests/test-edit-flow-notifications.php
cojennin: Ah yea, good catch

tests/test-edit-flow-notifications.php
cojennin: Hm, yea, not sure why this is part of these tests anyway, doesn't seem like it matters to have it?

tests/test-edit-flow-notifications.php
cojennin: Does this have a parent function?

They throw an exception when running PHPUnit: `Call to undefined method WP_UnitTestCase::wpSetUpBeforeClass()`

tests/test-edit-flow-notifications.php
cojennin: Same thing here, does this have a parent function?

It throws an exception when running PHPUnit: `Call to undefined method WP_UnitTestCase::wpSetUpBeforeClass()`

tests/test-edit-flow-notifications.php
mjangda: > would $factory be available in setUp?

Yep, it would.

I find it's generally safer and better to set up the test data on a per-test basis (wherever possible). It's a bit slower but not as risky (bugs in tests are super painful to debug).

I don't think it's a big deal here since we're not manipulating the user in any way, so I'd be okay with leaving it.

mjangda: Ah interesting (https://make.wordpress.org/core/handbook/testing/automated-testing/writing-phpunit-tests/#shared-fixtures):

> Sometimes, all tests in a class can use the same database fixtures. Generating these fixtures inside of setUp() will cause them to be recreated for each test, which can dramatically decrease test performance. In these cases, it’s possible to create fixtures once, before any tests in the class have run, using wpSetUpBeforeClass()

tests/test-edit-flow-notifications.php
mjangda: Huh, weird, I guess not!

tests/test-edit-flow-notifications.php
cojennin: Ah hey, good to know

modules/dist/notifications.build.js
```diff
@@ -0,0 +1,263 @@
+/******/ (function(modules) { // webpackBootstrap
```
mjangda: This file seems like it snuck in so going to delete.

PR #562, opened 2019-12-04
Dockerize and add e2e testing

cojennin:
Dockerizes testing on Travis (and locally) based on [Gutenberg](https://github.com/WordPress/gutenberg/blob/master/.travis.yml).

Adds an E2E test to confirm E2E testing works and the plugin is loaded.


cojennin:
The E2E test will continue to fail until #560 is merged in, the E2E tests [delete all posts before running the tests](https://github.com/Automattic/Edit-Flow/pull/562/files#diff-c05de4393aeadf1e2e3c30ea7fec0bf9R64), but the delete action is throwing a WSOD because of the notice from #560 (see screenshot) which prevents the E2E tests from completing (causing the failure)

Screen Shot 2019-12-04 at 9 51 10 AM

cojennin:
Includes a quick update to [some tests](https://github.com/Automattic/Edit-Flow/pull/562/files#diff-84e591ce27341a09991a958fab474ccdR340) that were causing test failures in PHPUnit

package.json
```diff
@@ -24,16 +44,22 @@
...
-    "sass-loader": "^7.1.0",
+    "lodash": "^4.17.15",
+    "node-sass": "^4.13.0",
+    "sass-loader": "^7.3.1",
     "webpack": "^4.27.1",
     "webpack-cli": "^3.1.2"
   },
-  "dependencies": {}
+  "dependencies": {
+    "@wordpress/e2e-test-utils": "^3.0.0",
```
mjangda: Shouldn't these be in `devDependencies`?

cojennin: Ah shoot, good catch

cojennin:
Made the small change @mjangda suggested (tiny tweak to package.json), approval got wiped out but should still be good to merge so we can move forward on adding e2e testing to PR's like #563 and #555

PR #564, opened 2019-12-09
Scope module assets for calendar

cojennin:
This is based on work from @justnorris in #487.

Instead of trying to tackle scoping assets across all of Edit Flow at once, we'll create a series of PR's to bring different modules into alignment and then introduce stricter guardrails once all the modules are in better shape.

### DIFF
--- a/common/php/class-module.php
+++ b/common/php/class-module.php
@@ -215,15 +215,15 @@ class EF_Module {
 	 */
 	function enqueue_datepicker_resources() {
 
-		// Add the first day of the week as an available variable to wp_head
-		echo "<script type=\"text/javascript\">var ef_week_first_day=\"" . get_option( 'start_of_week' ) . "\";</script>";
-
 		wp_enqueue_script( 'jquery-ui-datepicker' );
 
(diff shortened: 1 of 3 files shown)


common/php/class-module.php
```diff
@@ -215,15 +215,15 @@ function filter_posts_link( $slug, $post_type = 'post' ) {
 	 */
 	function enqueue_datepicker_resources() {
 
-		// Add the first day of the week as an available variable to wp_head
-		echo "<script type=\"text/javascript\">var ef_week_first_day=\"" . get_option( 'start_of_week' ) . "\";</script>";
```
cojennin: This line was making the PHPUnit tests around enqueuing noisy, so moved it to use `wp_add_inline_script`

WPprodigy:
Let's hold off on this one until we have a good pattern down in https://github.com/Automattic/Edit-Flow/pull/565.

Once the first is merged and a good pattern is setup, the rest can follow suite.

cojennin:
Closing this for now, going to rework based on thoughts in https://github.com/Automattic/Edit-Flow/pull/565#issuecomment-586378819

PR #565, opened 2019-12-09
Scope module assets for custom status

cojennin:
This is based on work from @justnorris in #487.

It continues the work of scoping module assets. It reintroduces the  `EF_Module_With_View` class from #487, but holds off on deprecating `is_whitelisted_functional_view` and `is_whitelisted_settings_view` (which will be deprecated once all modules have been migrated).

It also introduces a small change to `Block_Editor_Compatible`, enabling the [registering/de-registering of multiple actions for the same hook](https://github.com/Automattic/Edit-Flow/compare/scope-module-assets-custom-status?expand=1#diff-228291de10e40133681b1ff8414a9a54R26)

Also, holding off on altering any usages of `disable_custom_statuses_for_post_type` for this PR, as it [caused issues with the original PR](https://github.com/Automattic/Edit-Flow/pull/487#issuecomment-449443659)

### DIFF
--- a/common/php/trait-module-with-view.php
+++ b/common/php/trait-module-with-view.php
@@ -1,0 +1,109 @@
+<?php
+
+trait EF_Module_With_View {
+
+	/**
+	 * Whether or not the current page is an Edit Flow settings view (either main or module)
+	 * Determination is based on $pagenow, $_GET['page'], and the module's $settings_slug
+	 * If there's no module name specified, it will return true against all Edit Flow settings views
+	 *
+	 * @since 0.8.3
+	 *
+	 * @param string $slug (Optional) Module name to check against
+	 * @return bool true if is module settings view
+	 */
+	public function is_module_settings_view( $slug = false ) {
+		global $pagenow, $edit_flow;
+
+		// All of the settings views are based on admin.php and a $_GET['page'] parameter
+		if ( 'admin.php' !== $pagenow || ! isset( $_GET['page'] ) )
+			return false;
+
+		$settings_view_slugs = array();
+		// Load all of the modules that have a settings slug/ callback for the settings page
+		foreach ( $edit_flow->modules as $mod_name => $mod_data ) {
+			if ( isset( $mod_data->options->enabled ) 
+				&& 'on' === $mod_data->options->enabled 
+				&& $mod_data->configure_page_cb )
+				$settings_view_slugs[] = $mod_data->settings_slug;
+		}
+
+		// The current page better be in the array of registered settings view slugs
+		if ( empty( $settings_view_slugs ) || ! in_array( $_GET['page'], $settings_view_slugs ) ) {
+			return false;
+		}
+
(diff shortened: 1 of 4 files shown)


WPprodigy:
Multiple inheritance patterns like this could get pretty messy down the road.

If a set of related helper methods are needed for specific modules, what do you think about using a trait instead? https://www.php.net/manual/en/language.oop5.traits.php

cojennin:
@WPprodigy Sure, I don't have a strong stance on the OOP methodology. Traits seem less arduous so I'm good with that

WPprodigy:
Looking better. I have a couple of thoughts:

1) I'm not a big fan of the current GB compatibility implementation. Seems to obfuscate quite a bit with little advantage and very limited flexibility (as seen here with you needing to extend how it works). I think I'm going to look into removing some of the abstraction layers, and give Modules more power over how it supports the block editor. This should help ease the efforts here as well.

2) The decision process for where a module decides if a script should be enqueued or not is kind of all over the place. It would be nice if we used the `init()` method as a very high-level overview of the module. Instead of tons of hooks inside of it. In other words, more descriptive wrapping methods like this:

```
function init() {
  $this->register_custom_statuses_taxonomy();
  $this->configure_settings_page();
  $this->integrate_custom_statuses();
}
```

Then configure_settings_page() would contain secondary level conditionals to control output:

```
function configure_settings_page() {
  add_action( 'admin_init', array( $this, 'register_settings' ) );

  if ( $this->is_current_module_settings_view() ) {
    // Or just the direct hook here if it's small / simple logic.
    $this->enqueue_settings_scripts();
  }
}
```

After level 1 (`init()`), wrapper methods could be used sparingly / only when it makes sense.

The benefit to this type of cleanup will be that we can clearly read from the top down, and decisions on whether or not something needs to be loaded will be more obvious.

cojennin:
> The decision process for where a module decides if a script should be enqueued or not is kind of all over the place. It would be nice if we used the init() method as a very high-level overview of the module. Instead of tons of hooks inside of it.

@WPprodigy agreed, could use some TLC.

I think a series of PR's would make things easier. Can move and rearrange a little bit at the time, first by focusing on separating module _settings page stuff_ and _non-settings page_ module stuff.

So you'd end up with something like

```
function init() {
  $this->configure_module();
  $this->configure_module_settings_page();
}
```

and go from there (separating it out more as needed)

GaryJones:
Closing this PR as it's been stale for over 6 years and the codebase has evolved significantly since then.

Thank you to @cojennin and @justnorris for the work on this, and to @WPprodigy for the thoughtful review feedback. The core problem (asset scoping) is already solved in the current codebase — assets are properly scoped via `is_whitelisted_settings_view()` and `is_whitelisted_page()`.

The improvements this PR proposed (reusable trait, cleaner `init()` structure, renaming "whitelist" terminology) are nice-to-haves for code quality, but would require significant rebasing effort. If there's appetite for this kind of modernisation work in the future, we can open a fresh issue to track it.

PR #569, opened 2019-12-15
Support i18n in story budget date picker

cojennin:
The story budget date picker has the same issue fixed in #546. Fix the i18n issues to better support localization of the story budget.


cojennin:
Small change around i18n, tests look good

PR #571, opened 2019-12-18
Add eslint

### LINKED ISSUE #558: Add eslint support
To help us catch coding standards and other more critical issues. Plus eslines to help us transition.

mjangda:
Add eslint using Automattic/eslint-config-wpvip which we use for all VIP projects. Includes eslines to allow transitioning over.

However, we may want to consider switching this to use https://www.npmjs.com/package/@wordpress/eslint-plugin instead which aligns more closely with Block Editor development + e2e testing, etc.

Fixes #558 

## Steps to Test

Run `npm install` + `npm run lint` to make sure lint runs correctly. If there are no changes, should just see 1000s or warnings but no errors.

Travis should run lint for when PHPCS runs but should also pass.


cojennin:
Looks good!

cojennin:
`However, we may want to consider switching this to use https://www.npmjs.com/package/@wordpress/eslint-plugin instead which aligns more closely with Block Editor development + e2e testing, etc.`

Ah, missed this. I'm still down for merging it to see what kind of results we get, and then migrate to that plugin once we get e2e testing setup and more Block development going

Issue #523, opened 2019-10-04
Post Slug not editable until post is set to a non-custom post status


kjbenk:
When Gutenberg is enabled I am seeing an issue where the post slug field is not saved, specifically it is cleared out entirely, when a post is saved with a custom status. I have traced done this issue to this code - https://github.com/Automattic/Edit-Flow/blob/0.9/modules/custom-status/custom-status.php#L1395-L1431. It appears that since the `$_POST['post_name']` element is not set when editing a post in Gutenberg, the function always thinks the user did not set a custom slug, and therefore it clears the post slug.

I have done some rough testing and I believe this logic can be removed entirely since the custom post status API has been updated over the years to better support custom post slugs.

cojennin:
@kjbenk thanks for reporting this. I think #575 should fix it, and I've added enough testing to hopefully catch most of the corner cases.

When I removed that entire function (`fix_post_name`), the `post_name` was getting permanently set (which prevented the slug from updating if the user had not intentionally set it). If you can take a look at that PR, that'd be fantastic, could use a few more eyes

htdat:
@kjbenk - I've tested this issue in Gutenberg and it no longer happens. I believe  #575 has already fixed this issue. 

I am closing this issue now but if the issue is still there, feel free to reopen it.

PR #575, opened 2020-01-09
Fix slug behaviour

cojennin:
Another take on fixing slug issues in #523

Added a bunch more tests, both around `wp_insert_post` and the Rest API to give a bit more confidence to this PR.

### To Test

The following should all be captured in the automated tests, but for manually testing:

**Do this for posts and pages**

_In Classic Editor & In Block Editor_
1. Create a post, save it as a custom status
2. Update the post title
3. Confirm the post slug has been updated

_In Classic Editor & In Block Editor_
1. Create a post, save it as a custom status
2. Set the post slug and save the post
3. Update the post title and save the post
4. Confirm the slug has not changed

_In Classic Editor & In Block Editor_
1. Create a page, save it as a custom status
2. Update the page title
3. Confirm the page slug has been updated

*Repeat this but make the page a child page

_In Classic Editor & In Block Editor_
1. Create a page, save it as a custom status
2. Set the page slug and save the post
3. Update the page title and save the post
4. Confirm the slug has not changed

*Repeat this but make the page a child page


cojennin:
This PR is already being used by other folks, so if no one has anything they want to flag I'll merge as is

PR #576, opened 2020-01-09
Small change to phpcs-diff

cojennin:
The filtering of the diff `-- '*.php'` is resulting in an empty file. With the usage of `--extensions=php` [in PHPCS should](https://github.com/Automattic/Edit-Flow/compare/fix/phpcs-diff?expand=1#diff-219bf2a45ce55a542f916415298efd7eR11) we should be able to avoid having to figure out how to filter on the diff


cojennin:
This is a small one line change to improve the PHPCS diff testing for Travis, going to merge as impact is negligible.

Issue #471, opened 2018-05-30
Remove $_REQUEST['create-note'] reference


TheCrowned:
[This code](https://github.com/Automattic/Edit-Flow/blob/master/modules/dashboard/widgets/dashboard-notepad.php#L56):

```php
if ( $current_notepad
	&& self::notepad_post_type == $current_notepad->post_type
	&& ! isset ( $_REQUEST['create-note'] ) ) {
	$new_note['ID'] = $current_id;
	wp_update_post( $new_note );
} else {
	wp_insert_post( $new_note );
}
```

checks for `$_REQUEST['create-note']`, which is a leftover from b740dcf741ad35dc1b273e009738fc7c32898a33. The else simply never happens.

PR #578, opened 2020-01-16
Clean up dashboard-note logic

WPprodigy:
Fixes https://github.com/Automattic/Edit-Flow/issues/471

The else clause was actually necessary for the first time a note was created. It was a bit confusing though, so I've cleaned up the logic to make a single path of execution. `wp_insert_post()` will handle updating the post if the ID is passed.

## Testing

Checkout the PR and update the dashboard note. Can also delete the post from the database manually and then ensure you're still able to create a note.

### DIFF
--- a/modules/dashboard/widgets/dashboard-notepad.php
+++ b/modules/dashboard/widgets/dashboard-notepad.php
@@ -18,7 +18,7 @@ class EF_Dashboard_Notepad_Widget {
 		register_post_type( self::notepad_post_type, array(
 				'rewrite' => false,
 				'label' => __( 'Dashboard Note', 'edit-flow' )
-			) 
+			)
 		);
 
 		$this->edit_cap = apply_filters( 'ef_dashboard_notepad_edit_cap', $this->edit_cap );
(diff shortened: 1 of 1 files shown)


modules/dashboard/widgets/dashboard-notepad.php
```diff
@@ -34,31 +34,29 @@ public function init() {
...
-			);
-		if ( $current_notepad
-			&& self::notepad_post_type == $current_notepad->post_type
-			&& ! isset ( $_POST['create-note'] ) ) {
-			$new_note['ID'] = $current_id;
-			wp_update_post( $new_note );
-		} else {
-			wp_insert_post( $new_note );
+		$note_data = array(
+			'post_content' => wp_filter_nohtml_kses( $_POST['note'] ),
```
cojennin: This is getting picked up by [PHPCS](https://travis-ci.org/Automattic/Edit-Flow/jobs/637800481?utm_medium=notification&utm_source=github_status) (scroll to the bottom of the output), can you address?

modules/dashboard/widgets/dashboard-notepad.php
```diff
@@ -34,31 +34,29 @@ public function init() {
...
+			'post_type'    => self::notepad_post_type,
+			'post_status'  => 'draft',
+			'post_author'  => get_current_user_id(),
+		);
+
+		$existing_notepad = get_post( absint( $_POST['notepad-id'] ) );
```
cojennin: Also getting picked up by [PHPCS](This is getting picked up by [PHPCS](https://travis-ci.org/Automattic/Edit-Flow/jobs/637800481?utm_medium=notification&utm_source=github_status) (scroll to the bottom of the output), can you address?) (scroll to bottom of the output), can you address as well?

PR #579, opened 2020-01-16
Display custom statuses in post states

WPprodigy:
Closes https://github.com/Automattic/Edit-Flow/issues/395

Custom post statuses has enough hacks as-is, let's adhere to core APIs and standards when we can :)

Instead of adding an extra column to the post list that takes up valuable space, let's follow core's UI and display custom statuses just like other statuses are shown in WP.

For some context, here is what is done in WP core: https://github.com/WordPress/WordPress/blob/dd4d98a368ca1d617adfaaef9b2937c60464ac4a/wp-admin/includes/template.php#L2100-L2170. Of note, I followed the same rules by letting core continue to handle some statuses specially, and by not showing the status when a specific post status is filtered.

It will also smartly handle cases where EF has changed a core statuses name (like pending). Thanks to the use of array keys, it'll override the default status display name with our custom ones.

## Testing

Apply the patch and try out different post statuses, post types, etc.

_____

## Preview

post-states

### DIFF
--- a/modules/custom-status/custom-status.php
+++ b/modules/custom-status/custom-status.php
@@ -702,7 +702,7 @@ class EF_Custom_Status extends EF_Module {
 	 *
 	 * @return array $post_states
 	 */
-	function add_status_to_post_states( $post_states, $post ) {
+	public function add_status_to_post_states( $post_states, $post ) {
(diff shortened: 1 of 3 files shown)


cojennin:
Hey great! Thanks for carrying the torch on this, I was giving it a shot with #561 but got side tracked. Definitely makes more sense

PR #580, opened 2020-01-16
Cleanup block editor compatibility logic

WPprodigy:
First pass as what I was describing earlier. Open for discussion.

## Description

_No functionality was harmed in the making of this PR._

This is purely development focused, no behavioral changes to the plugin should occur. This same events happen when loading the plugin, the same conditionals are checked, scripts are enqueued the same way and time, etc.

There were three mains reasons for this change:

1. Modules should have more flexibility around how they integrate with Gutenberg, and being limited to a single unhooking/hooking process is quite restrictive. (metaboxes may also need to be hidden, etc)

2. The responsibility belongs to the module. The base Edit_Flow class shouldn't need to know about such details. A lot of extra abstractions were removed with the same end result here, though in my opinion it will create a much easier development flow / comprehension.

3. Gutenberg / Block Support should be a first class citizen moving forward, not shoved away in a corner. I'd prefer to see metaboxes get that treatment in the future 😄

There is more that can be done to clean up how a module reads and the decision tree on whether or not we are loading GB scripts, but I opted to keep this PR as simple as possible with the single purpose of breaking out of the abstraction layers. https://github.com/Automattic/Edit-Flow/pull/565 can take that part further.

There were also quite a bit of unused code/pieces that I removed. Version control has us covered if they are ever needed, though I don't think that will be the case. https://github.com/Automattic/Edit-Flow/commit/dde1d2dcbb1fc7ac3b21291f409bcd3d05b0719b is the main commit here. And as you can see, we get the same end behavior for much less code, without losing anything.

## Steps to Test

Checkout this PR and ensure that the Custom Statuses functionality works the same in Gutenberg as it did before.


cojennin:
Yea, makes sense. I was looking for more context around the original implementation. Not much there, looks like the reasoning was it meant [avoiding modifying the existing module code.](https://github.com/Automattic/Edit-Flow/blame/master/BLOCKS.md#L88) 

Can you resolve the conflict and the [two PHPCS issues](https://travis-ci.org/Automattic/Edit-Flow/jobs/638173279?utm_medium=notification&utm_source=github_status) (ignoring the indenting issues, the whole file is indented incorrectly). Then I'll test and approve if everything looks good

WPprodigy:
I'm pretty sure I've switched that same short array syntax error in the same spot twice now :shakesfirstatclouds:

At least maybe now there won't be a merge conflict.

Issue #583, opened 2020-01-22
Conflict with Gutenberg 7.2+


ryelle:
When trying to update Gutenberg on WordCamp.org, I noticed this issue with draft posts… The entire editor crashes 🙁  When I downgrade back to GB 7.1, everything's fine again.

![Screen Shot 2020-01-22 at 6 05 23 PM](https://user-images.githubusercontent.com/541093/72942581-caa1ee80-3d41-11ea-8773-63d715d3794b.png)

The error is this:

```
Error: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
```

<details>
<summary>Full stack trace (not actually useful, but for completeness)</summary>
<pre>
at ah (https://example.local/wp-content/plugins/gutenberg/vendor/react-dom.min.b694e242.js?ver=16.9.0:109:220)
    at Ti (https://example.local/wp-content/plugins/gutenberg/vendor/react-dom.min.b694e242.js?ver=16.9.0:145:151)
    at unstable_runWithPriority (https://example.local/wp-content/plugins/gutenberg/vendor/react.min.0212dc62.js?ver=16.9.0:26:340)
    at Ma (https://example.local/wp-content/plugins/gutenberg/vendor/react-dom.min.b694e242.js?ver=16.9.0:52:280)
    at Ia (https://example.local/wp-content/plugins/gutenberg/vendor/react-dom.min.b694e242.js?ver=16.9.0:140:407)
    at ze (https://example.local/wp-content/plugins/gutenberg/vendor/react-dom.min.b694e242.js?ver=16.9.0:118:14)
    at https://example.local/wp-content/plugins/gutenberg/vendor/react-dom.min.b694e242.js?ver=16.9.0:53:49
    at unstable_runWithPriority (https://example.local/wp-content/plugins/gutenberg/vendor/react.min.0212dc62.js?ver=16.9.0:26:340)
    at Ma (https://example.local/wp-content/plugins/gutenberg/vendor/react-dom.min.b694e242.js?ver=16.9.0:52:280)
    at mg (https://example.local/wp-content/plugins/gutenberg/vendor/react-dom.min.b694e242.js?ver=16.9.0:52:496)
    at V (https://example.local/wp-content/plugins/gutenberg/vendor/react-dom.min.b694e242.js?ver=16.9.0:52:431)
    at Sb (https://example.local/wp-content/plugins/gutenberg/vendor/react-dom.min.b694e242.js?ver=16.9.0:213:96)
[...]

PR #584, opened 2020-01-23
Fix custom status error with Gutenberg 7.2+

cojennin:
A fix for #583. Continuing to break the rules, but for a good cause. 

Props to @ryelle for tracking this down

## Steps to Test

1. With Gutenberg 7.2 or 7.3
2. With the custom status module active
3. Try to edit a draft post
4. Editor should not crash


ryelle:
Loading, editing, and saving a draft works, but after the draft finishes saving, the button reverts to "save draft" 

Screen Shot 2020-01-23 at 11 16 32 AMScreen Shot 2020-01-23 at 11 16 22 AM

I'm also seeing a crash with a custom status ("assigned")— this time, the post loads fine, but crashes when I try to save it. The error is the same: 

```
Error: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
```

WPprodigy:
Hmm, so GB is changes the markup pretty drastically during the saving states. Perhaps utilizing more of the selectors could help here, like `select( 'core/editor' ).isSavingPost()`. I'll see what I can find.

Also noting that https://github.com/WordPress/gutenberg/pull/12517 is another issue to watch as a better way to solve this.

WPprodigy:
Ugh, and there is also https://github.com/WordPress/gutenberg/issues/17632 that would help here (was closed in favor of https://github.com/WordPress/gutenberg/issues/13413, though it is a bit different IMO).

WPprodigy:
Tried a different approach in https://github.com/Automattic/Edit-Flow/pull/585

cojennin:
Closing in favor of #585

PR #585, opened 2020-01-24
Keep save button text updated, without errors.

WPprodigy:
Fixes https://github.com/Automattic/Edit-Flow/issues/583. A different approach to https://github.com/Automattic/Edit-Flow/pull/584.

### Fixing the page-breaking error
On the initial load, when the `subscribe()` method triggers we can update the button text right away from when it first exists. By checking specifically if the `innerText` is one of the two values we don't want, we avoid the errors that occur in https://github.com/Automattic/Edit-Flow/issues/583.

### Keeping the button text changes even after post saving
Then there is the issue of the button text reverting back to "Save draft" / "Save as Pending" after saving the post.

When a post is being saved, the DOM change quite a bit (4 different states) and there is unfortunately no way that I could find to hook in after the save is completed and the button is back to it's default state (the last animation tick being particularly annoying). To workaround this limitation, I'm using the MutationObserver API if the browser has it available: https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver

So then when the `subscribe()` method is triggered and the post is being saved, we set up a single observer on the parent element. This observer waits for the button to be added back then it targets the `innerText` again and get it updated. This node is actually re-added quite a few times per post-save, and due to various animations it's not actually possible to set/remove the observer each time. So instead, I decided to just set up one observer and keep it running. It watches over the minimal amount of nodes necessary.

### Testing
1. Install and activate Gutenberg plugin (7.3)
2. Ensure custom statuses module is enabled.
3. Create/edit posts, change around statuses, re-open old drafts, etc.
4. Make sure the button always just says "Save" rather than "Save Draft", etc.
5. Do the same tests without the GB plugin active, for BC testing.


WPprodigy:
I know there are new coding standards, but I'd prefer to keep this PR single purposed. We should clean up this file in a future PR.

cojennin:
@WPprodigy Looks good! I thought about using a [MutationObserver originally](https://github.com/Automattic/Edit-Flow/pull/555#issuecomment-558193745), but was waffling on IE support. I think it's probably fine.

If we're going to overhaul it this much I'd like to roll in [these changes](https://github.com/Automattic/Edit-Flow/pull/555/files#diff-228291de10e40133681b1ff8414a9a54R437) that support setting the correct label and [these e2e tests](https://github.com/Automattic/Edit-Flow/pull/555/files#diff-34271a6a12575809138dc7f55ed269a5R1) that validate it. If you can add those changes along with the e2e tests I'd feel confident enough to merge the change.

cojennin:
Also, add a `.eslintignore` to the top level of the project with `*.build.js`, that should eliminate most of the warnings in Travis

blocks/src/custom-status/block.js
```diff
@@ -14,47 +14,56 @@ let { SelectControl } = wp.components;
...
-
-  // For new posts, we need to force the our default custom status.
-  // Otherwise WordPress will force it to "Draft".
-  const isCleanNewPost = select( 'core/editor' ).isCleanNewPost();
-  if ( isCleanNewPost ) {
-    dispatch( 'core/editor' ).editPost( {
-      status: ef_default_custom_status
-    } );
-
-    return;
```
cojennin: Not sure why this `return` was here in the first place, but why drop it?

WPprodigy:
@cojennin Could we do the eslint / standards fixing in a followup PR right after this? And then after that, a new PR for the feature of adding the correct labels into the button can follow.

I see a few PRs now have tried the above, would prefer to do these in piece mail so they stop getting left behind :). That way we can get this out the door, and it's more visible what happened when.

And then agreed, I'll get some comments added, perhaps just linking back to the issue for now to have the context and make it easier to remove in the future.

blocks/src/custom-status/block.js
WPprodigy: I'll do some testing around this, but logically felt like it should be removed in case the button needs updating / an observer adding during that subscribe().

cojennin:
@WPprodigy Sure, let's move forward on merging this PR. For future PR's I'd like to get in the habit of incorporating e2e tests where possible for frontend changes, just to make sure we're not regressing (as @mjangda cautioned in #555, want to avoid things like [infinite loops](https://github.com/Automattic/Edit-Flow/pull/555#discussion_r349986067) or observers)

blocks/src/custom-status/block.js
WPprodigy: Doesn't seem to matter really - this dispatch is sent immediately when a new post is loaded, before the save button is visible.

If core were to show the save button, but perhaps disabled, then we would want the logic to "fallthrough" like it currently does. So the current implementation I guess is just a little more prepared if anything, but ultimately doesn't matter much since the subsequent checks are very fast, and they are 10 bajillion subscribe() calls anyways 😆

PR #587, opened 2020-01-27
Fix ef_custom_status_list filter on get_custom_statuses()

### LINKED ISSUE #586: ef_custom_status_list filter doesn't seem to be working
I've copied the code *exactly* from the ["Limit custom statuses based on user role"](https://editflow.org/extend/limit-custom-statuses-based-on-user-role/) page, then deleted all `permitted_statuses` except `draft`, and the `contributor` role is still able to set any custom status.

If I `print_r( $custom_statuses )` right before doing the `return`, I'm seeing that `$custom_statuses` does have the correct data, but just doesn't seem to be affecting anything through the filter.

jsit:
## Description

This fixes #586, where filtering custom statuses by user role wasn't working.

## Steps to Test

1. Check out PR.
1. Limit user role to subset of custom statuses as described [here](https://editflow.org/extend/limit-custom-statuses-based-on-user-role/)
1. Click the custom status drop-down while editing/creating a post as that user role, and see which custom statuses you have access to


cojennin:
Ah yea, good spot @jsit. Can you apply the filter [here](https://github.com/Automattic/Edit-Flow/blob/6b88e56bb421271cddc93c6df975c8fd3abfba48/modules/custom-status/custom-status.php#L294) instead?

That makes it consistent with it's other [usage](https://github.com/Automattic/Edit-Flow/blob/6b88e56bb421271cddc93c6df975c8fd3abfba48/modules/custom-status/custom-status.php#L405).

You probably put the change where we'll want to put it eventually, but I'd like some more testing around it before we make it permanent. It'll be less impactful to put it [here](https://github.com/Automattic/Edit-Flow/blob/6b88e56bb421271cddc93c6df975c8fd3abfba48/modules/custom-status/custom-status.php#L294) and then hopefully we can get it out with the next release in a few days (*fingers crossed*)

jsit:
Thanks @cojennin; do the changes I've just committed make sense?

WPprodigy:
> You probably put the change where we'll want to put it eventually, but I'd like some more testing around it before we make it permanent.

I agree. And on a quick look, I can see a few potential pit falls to putting it there without testing. It definitely should have been there from the start though 😣

> do the changes I've just committed make sense?

Looks good @jsit ! Just tested and works as expected.

Issue #588, opened 2020-01-28
Causes duplicate custom fields to be created


jsit:
## Steps to Reproduce the Problem

1. Activate the Edit Flow plugin
1. Add a Custom Field to a post
1. See that the Custom Field is added twice

<img width="965" alt="Screen Shot 2020-01-28 at 12 10 58 PM" src="https://user-images.githubusercontent.com/643417/73287638-8bebb880-41c7-11ea-8665-efc428ba5049.png">

cojennin:
@jsit Hm, that's a problem. #591 should hopefully fix

cojennin:
Fixed in Edit Flow 0.9.5

PR #591, opened 2020-01-29
Prevent multiple custom fields from being created

cojennin:
A fix for #588

For some reason, when `post` is a dependency of the Editorial Comment js, the custom fields get duplicated. Not sure why we have `post` as a dependency in the first place since all we use is `wpAjax`. 

Also, created a quick e2e test to validate the creation of an editorial comment


cojennin:
e2e tests look good, going to merge

PR #590, opened 2020-01-28
Move 'Customize' link in Metadata metabox to inside metabox

jsit:
Fixes #589

PR #592, opened 2020-01-30
Add back escape

cojennin:
Spotted this just after I hit the merge button on #590 

The `esc_url` was dropped, adding it back in


cojennin:
One liner with low impact, going to merge (apologies for not spotting this earlier)

PR #602, opened 2020-04-08
Fixing up PHPCS errors for calendar

cojennin:
Fixing up all the existing PHPCS errors for `calendar.php`.


cojennin:
Just cleaning up some PHPCS errors for the calendar

PR #614, opened 2020-05-02
Allow scheduled posts to be shifted around on calendar

cojennin:
The calendar doesn't support moving `published_status` posts. Up to this point, scheduled posts have been considered published. This PR changes it so that scheduled posts are no longer considered "published" and can be moved around the calendar.


cojennin:
This includes some extra E2E testing around the drag and drop of the calendar, to better test support for published, unpublished and scheduled posts

Issue #630, opened 2021-04-12
Travis tests fail


htdat:
Travis tests in the recent PRs fail https://github.com/Automattic/Edit-Flow/pulls?q=is%3Apr+ but they worked well before. 

![Markup on 2021-04-12 at 11:20:54](https://user-images.githubusercontent.com/10045087/114340176-2b089000-9b81-11eb-96d8-2308460ecf1d.png)

## Detailed errors 

- Oldest https://travis-ci.org/github/Automattic/Edit-Flow/jobs/743803874
```
gyp ERR! build error 
gyp ERR! stack Error: `make` failed with exit code: 2
gyp ERR! stack     at ChildProcess.onExit (/home/travis/build/Automattic/Edit-Flow/node_modules/node-gyp/lib/build.js:262:23)
gyp ERR! stack     at ChildProcess.emit (node:events:329:20)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (node:internal/child_process:277:12)
gyp ERR! System Linux 4.4.0-101-generic
gyp ERR! command "/home/travis/.nvm/versions/node/v15.2.0/bin/node" "/home/travis/build/Automattic/Edit-Flow/node_modules/node-gyp/bin/node-gyp.js" "rebuild" "--verbose" "--libsass_ext=" "--libsass_cflags=" "--libsass_ldflags=" "--libsass_library="
gyp ERR! cwd /home/travis/build/Automattic/Edit-Flow/node_modules/node-sass
gyp ERR! node -v v15.2.0
gyp ERR! node-gyp -v v3.8.0
gyp ERR! not ok 
Build failed with error code: 1
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! node-sass@4.13.0 postinstall: `node scripts/build.js`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the node-sass@4.13.0 postinstall script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
``` 
- Latest https://travis-ci.org/github/Automattic/Edit-Flow/jobs/762039246

First errors 

```bash
npm ERR! code 1
npm ERR! path /home/travis/build/Automattic/Edit-Flow/node_modules/node-sass
npm ERR! command failed
```

htdat:
Closing this issue, and deferring to https://github.com/Automattic/Edit-Flow/issues/634

PR #635, opened 2021-04-22
Move from Travis CI to GitHub Actions

### LINKED ISSUE #634: Migrating Travis CI to GitHub Actions
As the title, we consider moving from Travis CI to GitHub Actions as https://travis-ci.org/ will be moved to http://www.travis-ci.com/ soon. 

Good start https://docs.github.com/en/actions/learn-github-actions/migrating-from-travis-ci-to-github-actions

Some examples: 
- https://github.com/Parsely/wp-parsely/tree/develop/.github 
- https://github.com/GaryJones/plugin-boilerplate/tree/master/.github

Working branch: https://github.com/Automattic/Edit-Flow/compare/add/gh-actions

Other things to consider: 

- [x] Convert Travis CI file to GitHub Actions.
- [x] Support PHP 8 (maybe 8.1 too)
- [x] Multisite

htdat:
Fix #634 

## Summary

As mentioned in the title, the main goal of this PR is to move all tests/checks from Travis CI (aka[ .travis.yml ](https://github.com/Automattic/Edit-Flow/blob/2e07a1b7b7af1e783b858700390da85b418c2405/.travis.yml)file) to GitHub Actions. 

I will not add new things (tests, checks, deploy actions) in this PR.

## What we're running with Travis-CI

From here
https://github.com/Automattic/Edit-Flow/blob/2e07a1b7b7af1e783b858700390da85b418c2405/.travis.yml#L94-L102

Travis-CI is running these tests/checks: 

1. PHP Unit tests 
2. E2E tests
3. PHPCS Diff
4. Lint (JS) 
5. Jest (JS Unit)

(1) is run for all matrix cases.
(2) to (5) are run for selected matrix cases, PHP 7.4 + WP 5.4 or latest (Note: the last commit to `travis.yml` was one year ago). 

## What this PR does 

After many trial-and-error attempts, I decided to separate two tests into workflows in GitHub actions: 

- E2E and JS tests: (2), (4) and (5) above.  Basically, these tests require `npm`. 
- PHP tests: (1) and (3) above - these tests require `composer`. 

###  E2E and JS tests

Basically, I am following up this set up https://github.com/WordPress/gutenberg/blob/trunk/.github/workflows/end2end-test.yml
But with some differences: 

- Switch to use `@wordpress/env` package as mentioned here https://www.npmjs.com/package/@wordpress/scripts/v/9.1.0#env-deprecated
- With this new `env` package, I can remove tons of [set-up code from travis.yml](http://github.com/Automattic/Edit-Flow/blob/2e07a1b7b7af1e783b858700390da85b418c2405/.travis.yml#L52-L88) 

### PHP tests

The way I am setting up this file is similar to the WP Parsely plugin https://github.com/Parsely/wp-parsely/blob/d7a06bd696a7af3e31e49ee98f893ff8a1d07755/.github/workflows/integration-tests.yml

I don't follow the Dockerize way like `E2E and JS tests` mainly due to a similar reason as mentioned here https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3691#issuecomment-761888298

Another note here is that I am using `matrix.wp` value to run tests for two latest WP versions 5.6 and 5.7.

## Steps to Test

Well, I am not sure if there is any way to tests the test runs except looking at workflow files and action results:

- https://github.com/Automattic/Edit-Flow/actions/runs/773363260 - E2E tests fail - actually, it's not an issue with GitHub actions. Instead, we need to fix this issue in our code. Here is the issue https://github.com/Automattic/Edit-Flow/issues/637
- https://github.com/Automattic/Edit-Flow/actions/runs/773363262 - Tests fail with PHP 8.0 and 8.1 - yeah, we need to fix the issue for our composer packages. Here is the issue https://github.com/Automattic/Edit-Flow/issues/638


composer.json
```diff
@@ -10,11 +10,22 @@
...
+        "composer/installers": "~1.0",
+        "php": ">=5.6"
     },
     "require-dev": {
         "automattic/vipwpcs": "^2.0",
-        "dealerdirect/phpcodesniffer-composer-installer": "^0.5.0",
-        "exussum12/coverage-checker": "^0.11.2"
+        "dealerdirect/phpcodesniffer-composer-installer": "^0.4.1 || ^0.5 || ^0.6.2 || ^0.7",
+        "exussum12/coverage-checker": "^0.11.2",
+        "phpunit/phpunit": "^4 || ^5 || ^6 || ^7"
```
htdat, line 20: This change and removing composer.json help handle different PHPUnit and PHP versions.

mikeyarce:
This is looking great, @htdat !  

For the PHP tests, can we allow failures for PHP > 8.0 for now?    Maybe something like `continue-on-error: true`?

What do you think about splitting this into two PRs - one for the PHP tests and one for the E2E tests?  I feel like we'll be able to merge the PHP ones pretty soon but the E2E ones might take a bit longer.

htdat:
@mikeyarce thanks for your review :D 

> For the PHP tests, can we allow failures for PHP > 8.0 for now? Maybe something like continue-on-error: true?

👍  My recent commit has addressed this https://github.com/Automattic/Edit-Flow/pull/635/commits/b45c8f5d4cf6151f9d3811f451c610d8fc23e3d1

> What do you think about splitting this into two PRs - one for the PHP tests and one for the E2E tests? I feel like we'll be able to merge the PHP ones pretty soon but the E2E ones might take a bit longer.

Yeah, I can for sure. However, IMHO, we might still want to keep those two tests in this PR. 
The main purpose of this PR is to move away from Travis CI and ensure that all tests in Travis configuration can run in GitHub Actions with minimal changes. 

In case, we need to upgrade composer and/or npm packages, add coverage, etc, we can do that in the next PRs. I've created a few issues for these: #636, #637, #638.

What do you think?

pkevan:
I don't see this mentioned but what is the reason for moving? Will there be anything to consider when running these tests with Github actions?

.github/workflows/php-tests.yml
```diff
@@ -0,0 +1,69 @@
...
+      WP_VERSION: ${{ matrix.wp }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        php: [ '5.6', '7.0', '7.1', '7.2', '7.3', '7.4' ]
+        wp: [ '5.6', '5.7' ]
+        allowed_failure: [ false ]
+        include:
+          - php: '8.0'
```
nielslange, line 24: Just wondering why WP 5.6 and WP 5.7 under PHP 8.0 are allowed to fail, given that PHP 8.0 is the current stable version of PHP.

.github/workflows/php-tests.yml
```diff
@@ -0,0 +1,69 @@
...
+        if: ${{ matrix.php >= 8.0 }}
+        uses: ramsey/composer-install@v1
+        with:
+          composer-options: --ignore-platform-reqs
+
+      - name: Start MySQL service
+        run: sudo systemctl start mysql.service
+
+      - name: Install WordPress test site
+        run: bash bin/install-wp-tests.sh wordpress_test root root localhost
```
nielslange, line 60: In line 21, I see that different WP versions being specified. However, `bin/install-wp-tests.sh wordpress_test root root localhost` would always install the latest WP version. I'd change that line to `bin/install-wp-tests.sh wordpress_test root root localhost ${{ matrix.wp }}` to install different WP versions for the tests.

nielslange (commented):
Looking good overall. I just noticed two spots for which I left a comment.

NIT: I also noticed that e2e-and-js-tests.yml, php-tests.yml and composer.json are missing an empty line by the end of the file.

htdat:
@pkevan - thanks for your great questions. 

> I don't see this mentioned but what is the reason for moving? 

The main reason is that Travis-CI.org will be shut down soon. 
And recently Travis CI tests have failed without any clear reason as I mentioned here https://github.com/Automattic/Edit-Flow/issues/630 But maybe something may change and I could not figure it out.

Then @mikeyarce suggested moving to GitHub Actions. 

When working and researching more with GitHub Actions, I notice that WordPress core and Gutenberg projects moved from Travis-CI.org to GitHub Actions too. With the fact that our current Travis-CI setup here based on Gutenberg setup (https://github.com/Automattic/Edit-Flow/pull/562), I think this makes sense for us to do the same.

> Will there be anything to consider when running these tests with Github actions?

I have not seen anything like that so far. Instead, it takes time to understand the differences between GitHub Actions and Travis, and follow the new way of Gutenberg in setting up testing environments (I mentioned details in **E2E and JS tests** section of this PR description).

htdat:
> NIT: I also noticed that e2e-and-js-tests.yml, php-tests.yml and composer.json are missing an empty line by the end of the file.

@nielslange - missing an empty line by the end of the file can cause any specific issue for .yml files? I know that might cause issues for other file types in some cases but I have done a bit of research and can not find the "why" for YAML files. 
I do not mean we need to follow WP core closely but even its GitHub Actions config files like [this](https://github.com/WordPress/wordpress-develop/blob/master/.github/workflows/javascript-tests.yml) also does not have any empty line by the end of the file.

.github/workflows/php-tests.yml
htdat: @nielslange - yeah, @mikeyarce suggested it here https://github.com/Automattic/Edit-Flow/pull/635#issuecomment-825211538 and I myself think that makes sense until this issue is fixed: https://github.com/Automattic/Edit-Flow/issues/638

.github/workflows/php-tests.yml
htdat: Thanks @nielslange. You're correct. I addressed this issue on 8688a81

htdat:
I've added this line into `PHP tests` section in the PR description above. 

> **Update on 2021.04.26:** Another reason is that wp-env has not supported multisite yet https://github.com/WordPress/gutenberg/issues/27961

mikeyarce (approved):
LGTM!

PR #643, opened 2021-05-11
PHP 8.0 and PHPCS: partly fix compatibility with composer packages

htdat:
Partly fix #638
Close #647

## Description

- Update "coverage-checker" so that it can support PHP 8.0.
- Stop running PHPCS for PHP 8.0 as WPCS is not ready as mentioned here https://github.com/Automattic/Edit-Flow/issues/638#issuecomment-825511819.
- Update bin/phpcs-diff.sh then (1) it will not fetch all branches (we just need `master` branch), and (2) it will show errors when diff lines do not conform with PHPCS. Currently, it does not exit with any error. 
- Remove tests for PHP 8.1 - [see here](https://github.com/Parsely/wp-parsely/blob/152c1dd9b288f9ae66e2c521171194df5f7170b8/.github/workflows/integration-tests.yml#L33).

## Steps to Test

Mainly check the relevant GitHub Actions for PHP Tests here. 

Note: for PHP 8.0 test failure, it will be addressed in another PR as the solution is here https://github.com/Automattic/Edit-Flow/issues/639#issuecomment-836384914

PR #644, opened 2021-05-12
Fix test_story_budget_set_number_days_filter_invalid in PHP 8.0

### LINKED ISSUE #639: PHP 8.0 Plugin Compatibility - test_story_budget_set_number_days_filter_invalid
## Expected/Desired Behavior

No issue with PHP 8.0 in the plugin.

## Actual Behavior

This test fail can be seen here https://github.com/htdat/Edit-Flow/runs/2434962761
When running PHPUnit tests, this is the error: 

```
> @php ./vendor/bin/phpunit --testsuite WP_Tests
  
Installing...
Running as single site... To run multisite, use -c tests/phpunit/multisite.xml
Not running ajax tests. To execute these, use --group ajax.
Not running ms-files tests. To execute these, use --group ms-files.
Not running external-http tests. To execute these, use --group external-http.
PHPUnit 7.5.20 by Sebastian Bergmann and contributors.

.......................................................F...       59 / 59 (100%)

Time: 2.77 seconds, Memory: 32.00 MB

There was 1 failure:

1) WP_Test_Edit_Flow_Story_Budget::test_story_budget_set_number_days_filter_invalid
Failed asserting that 'not days' matches expected 1.

/home/runner/work/Edit-Flow/Edit-Flow/tests/test-edit-flow-story-budget.php:92
``` 

Tests fail with: 
- PHP 8.0 and WP 5.6
- PHP 8.0 and WP 5.7

htdat:
Fix #639 

## Description

Details here https://github.com/Automattic/Edit-Flow/issues/639#issuecomment-836384914
This PR should be merged after this PR https://github.com/Automattic/Edit-Flow/pull/643 - actually, this PR is rebased on PR 643. 

## Steps to Test

1. Check out PR.
1. Set up WP test site `bash bin/install-wp-tests.sh wordpress_test root root`.
2. Run PHPUnit tests in PHP 8.0.x `composer integration`.
3. Confirm that all tests pass.

### DIFF
--- a/modules/story-budget/story-budget.php
+++ b/modules/story-budget/story-budget.php
@@ -229,6 +229,7 @@ class EF_Story_Budget extends EF_Module {
 		}
 
 		if ( isset( $new_filters['number_days'] ) ) {
+			$new_filters['number_days'] = intval( $new_filters['number_days'] );
 			if ( $new_filters['number_days'] <= 1 ) {
 				$existing_filters['number_days'] = 1;
 			} else {


mikeyarce:
Looking at the code, it looks like this is set to be `(int)` here: 
https://github.com/Automattic/Edit-Flow/blob/124c3cde900c22abe15c04b0b2bd11e306b2aef4/modules/story-budget/story-budget.php#L201

However, in the test it doesn't look like we're casting as `(int)` here:
https://github.com/Automattic/Edit-Flow/blob/2e07a1b7b7af1e783b858700390da85b418c2405/tests/test-edit-flow-story-budget.php#L68

Instead of changing the code, would it make more sense to update the test to cast as `(int)`?

htdat:
@mikeyarce 

> Looking at the code, it looks like this is set to be `(int)` here: 
https://github.com/Automattic/Edit-Flow/blob/124c3cde900c22abe15c04b0b2bd11e306b2aef4/modules/story-budget/story-budget.php#L201

This is a good point!

> However, in the test it doesn't look like we're casting as `(int)` here:
https://github.com/Automattic/Edit-Flow/blob/2e07a1b7b7af1e783b858700390da85b418c2405/tests/test-edit-flow-story-budget.php#L68
> Instead of changing the code, would it make more sense to update the test to cast as `(int)`?

L68 here is already an `int` so I don't think we need to cast it. 

However, actually, #639 mentions the error in other lines and test case: 
https://github.com/Automattic/Edit-Flow/blob/2e07a1b7b7af1e783b858700390da85b418c2405/tests/test-edit-flow-story-budget.php#L88-L92

We can get rid of this test case but I think it's still good to keep this test and practice defensive coding for the `EF_Story_Budget::update_user_filters_from_form_date_range_change` method as it can be used somewhere else in the future?

mikeyarce (approved):
After discussion, this sounds like a good approach!

PR #650, opened 2021-05-19
Fix: Give each attribute in Edit Flow settings page a different id

### LINKED ISSUE #631: Error: Found 8 elements with non-unique id #change-module-nonce
## Expected/Desired Behavior

No error 

## Actual Behavior

Visit /wp-admin/admin.php?page=ef-settings and see this error in the browser console: 

```
Found 8 elements with non-unique id #change-module-nonce:
<input type="hidden" id="change-module-nonce" name="change-module-nonce" value="99ceb67ba4">
``` 

Relevant code
https://github.com/Automattic/Edit-Flow/blob/2e07a1b7b7af1e783b858700390da85b418c2405/modules/settings/settings.php#L272

htdat:
Fix #631 

---

## Description

This PR is pretty simple. It's simply to use `slug` of each module and append it to the current name/id. 

## Steps to Test

1. Apply this PR.
2. Visit wp-admin/admin.php?page=ef-settings
3. No longer see `Found 8 elements with non-unique id #change-module-nonce` error in your browser console.
4. See this page's HTML source, search `change-module-nonce`, there will be 8 results with a different suffix.

### DIFF
--- a/modules/settings/settings.php
+++ b/modules/settings/settings.php
@@ -269,7 +269,7 @@ class EF_Settings extends EF_Module {
 				if ( $mod_data->options->enabled == 'off' ) echo ' style="display:none;"';				
 				echo ' value="' . __( 'Disable', 'edit-flow' ) . '" />';
 				echo '</p>';
-				wp_nonce_field( 'change-edit-flow-module-nonce', 'change-module-nonce', false );
+				wp_nonce_field( 'change-edit-flow-module-nonce', 'change-module-nonce-' . $mod_data->slug, false );
 				echo '</form>';
 				echo '</div>';
 			}
--- a/modules/settings/lib/settings.js
+++ b/modules/settings/lib/settings.js
@@ -26,7 +26,7 @@ jQuery(document).ready(function(){
 			var module_action = 'disable';
 		
 		var slug = jQuery(this).closest('.edit-flow-module').attr('id');
-		var change_module_nonce = jQuery('#' + slug + ' #change-module-nonce').val();
+		var change_module_nonce = jQuery('#' + slug + ' #change-module-nonce-' + slug).val();
 		var data = {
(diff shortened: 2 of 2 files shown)


modules/settings/settings.php
```diff
@@ -269,7 +269,7 @@ function print_modules() {
 				if ( $mod_data->options->enabled == 'off' ) echo ' style="display:none;"';				
 				echo ' value="' . __( 'Disable', 'edit-flow' ) . '" />';
 				echo '</p>';
-				wp_nonce_field( 'change-edit-flow-module-nonce', 'change-module-nonce', false );
+				wp_nonce_field( 'change-edit-flow-module-nonce', 'change-module-nonce-' . $mod_data->slug, false );
```
natebot, line 272: What does this name change mean for [validating the nonce](https://github.com/Automattic/Edit-Flow/blob/124c3cde900c22abe15c04b0b2bd11e306b2aef4/modules/settings/settings.php#L100)?

natebot (changes requested):
Left a note asking about the nonce validation.

modules/settings/settings.php
natebot: nm, I see the nonce doesn't change based on name.

natebot (approved):

modules/settings/settings.php
htdat: > nm, I see the nonce doesn't change based on name.

Yeah, that's right. The `change-edit-flow-module-nonce` action for [wp_nonce_field](https://developer.wordpress.org/reference/functions/wp_nonce_field/) and [wp_verify_nonce](https://developer.wordpress.org/reference/functions/wp_verify_nonce/) is still the same.

htdat:
Thanks @natebot :D

PR #653, opened 2021-06-04
add 'cs' script to composer for easy running of phpcs

trepmal:
Enables `composer cs` for easily running PHPCS over project


htdat:
LGTM! 
It's working well when running with `composer cs` and `composer cbf`.

htdat (approved):

nielslange (approved):
LGTM as well.

Issue #654, opened 2021-06-07
Bug:  Quick Edit in User Groups PHP Error


mikeyarce:
## Expected/Desired Behavior

No PHP Errors when using Quick Edit

## Actual Behavior

PHP Error:

```
[07-Jun-2021 17:34:31 UTC] PHP Warning:  Undefined array key "hook_suffix" in /wp-admin/includes/class-wp-screen.php on line 223
```
In the console:

```
component: "Plugin: Edit-Flow"
file: "wp-admin/includes/class-wp-screen.php"
key: "d5841370c9175d9d14dff829bd2b4b69"
line: 224
message: "Undefined array key \"hook_suffix\""
stack: Array(6)
0: "WP_Screen::get()"
1: "convert_to_screen()"
2: "WP_List_Table->__construct()"
3: "EF_Usergroups_List_Table->__construct()"
4: "EF_User_Groups->handle_ajax_inline_save_usergroup()"
5: "do_action('wp_ajax_inline_save_usergroup')"
length: 6
__proto__: Array(0)
type: "warning"
__proto__: Object
```

## Steps to Reproduce the Problem

1. Go to Edit Flow -> User Groups 
2. Click Quick Edit
3. Make a change
4. Click "Update User Group"

## (Optional) Additional notes

I think the problem is here: https://github.com/Automattic/Edit-Flow/blob/master/modules/user-groups/user-groups.php#L1097-L1103

The `__construct()` should have `screen` because the method WordPress uses to determine this doesn't run when an AJAX call is made.

htdat:
As mentioned in https://github.com/Automattic/Edit-Flow/pull/666#pullrequestreview-681493522, this is a bug in core WP
https://core.trac.wordpress.org/ticket/49089

htdat:
## Replicate this issue in the core 

 - Add this file under wp-content/mu-plugins https://gist.github.com/htdat/06d77ea326d53cb1ae1e6b68c96aa0c3
 - Visit wp-admin/edit.php
 - Try to edit a post with "Quick Edit" to trigger the ajax action
 - Check PHP error log (debug.log)

Error with track trace: 

>  PHP Notice:  Undefined index: hook_suffix in wp-admin/includes/class-wp-screen.php on line 223
>  PHP Stack trace:
>  1. {main}() .composer/vendor/laravel/valet/server.php:0
>  2. require() .composer/vendor/laravel/valet/server.php:219
[...]

PR #666, opened 2021-06-10
Fix undefined hook_suffix

mikeyarce:
## Description

Fixes https://github.com/Automattic/Edit-Flow/issues/654

Without a specified `screen` when extending `WP_List_Table` - the function `convert_to_screen` will return a PHP Warning due to an undefined hook suffix which is needed to identify the current screen

## Steps to Test

1. Go to Edit Flow -> User Groups
2. Click Quick Edit
3. Make a change
4. Click "Update User Group"
5. You should see a PHP Warning like this:  `[07-Jun-2021 17:34:31 UTC] PHP Warning:  Undefined array key "hook_suffix" in /wp-admin/includes/class-wp-screen.php on line 223`

Apply this PR and then do the steps again and you won't see the PHP Warning any longer.

### DIFF
--- a/modules/custom-status/custom-status.php
+++ b/modules/custom-status/custom-status.php
@@ -1730,7 +1730,7 @@ class EF_Custom_Status_List_Table extends WP_List_Table
 			'plural' => 'custom statuses',
 			'singular' => 'custom status',
 			'ajax' => true,
-            'screen' => 'ef-custom-status-settings'
+            'screen' => 'ef-custom-status-settings',
 		) );
 
 	}
--- a/modules/custom-status/custom-status.php
+++ b/modules/custom-status/custom-status.php
@@ -1730,7 +1730,7 @@ class EF_Custom_Status_List_Table extends WP_List_Table
 			'plural' => 'custom statuses',
 			'singular' => 'custom status',
 			'ajax' => true,
-            'screen' => 'ef-custom-status-settings',
+			'screen' => 'ef-custom-status-settings',
 		) );
 
 	}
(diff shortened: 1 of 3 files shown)


modules/custom-status/custom-status.php
```diff
@@ -1729,7 +1729,8 @@ function __construct() {
 		parent::__construct( array(
 			'plural' => 'custom statuses',
 			'singular' => 'custom status',
-			'ajax' => true
+			'ajax' => true,
+			'screen' => 'ef-custom-status-settings',
```
htdat, line 1733: Actually, any value different from `null` will get you through this error.

htdat (approved):
I dug into this issue further, and I found out that there is nothing wrong with the current implementation of Edit Flow. 

It turns out the error itself is a bug of WordPress core. 
https://core.trac.wordpress.org/ticket/49089

That said, this PR helps overcome the core bug above.

mikeyarce:
> I dug into this issue further, and I found out that there is nothing wrong with the current implementation of Edit Flow.
> 
> It turns out the error itself is a bug of WordPress core.
> https://core.trac.wordpress.org/ticket/49089
> 
> That said, this PR helps overcome the core bug above.

Good find!

I saw that adding the `screen` argument was how this was addressed in the past with similar issues:
https://core.trac.wordpress.org/changeset/22085

Do you think there's anything we can suggest to Core that might improve this so that you don't need the `screen` argument?  It seems like it might only be an issue for AJAX requests.

htdat:
> I saw that adding the `screen` argument was how this was addressed in the past with similar issues:
https://core.trac.wordpress.org/changeset/22085

I am pretty sure that it does not help to fix the issue as WP_List_Table class can be still initialized with `screen = null`, and the issue is still there. 

> Do you think there's anything we can suggest to Core that might improve this so that you don't need the `screen` argument?  It seems like it might only be an issue for AJAX requests.

Yeah, I tried to find a simple way to replicate this issue last week but I failed somehow. I think this week with a fresh mind helps me to understand the root cause. I am going to write it down in #654 then post it to the trac issue.

rebeccahum:
Closing since it looks like it's been resolved per the trac ticket.

Issue #656, opened 2021-06-07
PHP Notice: Only variables should be passed by reference


mikeyarce:
## Expected/Desired Behavior

No PHP Notices.

## Actual Behavior

PHP Notice: 

```
[07-Jun-2021 18:36:16 UTC] PHP Notice:  Only variables should be passed by reference in wp-content/plugins/Edit-Flow/common/php/screen-options.php on line 169
```

```
component: "Plugin: Edit-Flow"
file: "wp-content/plugins/Edit-Flow/common/php/screen-options.php"
key: "b70f237b117c737f6e624ed611138805"
line: 169
message: "Only variables should be passed by reference"
stack: Array(2)
0: "wsScreenOptions10->ajax_save_callback()"
1: "do_action('wp_ajax_save_settings-ef_story_budget_screen_columns')"
length: 2
__proto__: Array(0)
type: "notice"
__proto__: Object
```

How does the plugin/feature/module currently work.

## Steps to Reproduce the Problem

1. Go to Dashboard > Story Budget
2. From the top click Screen
3. Change "Number of Columns" to something else

## (Optional) Additional notes

The problem comes from: https://github.com/Automattic/Edit-Flow/blob/master/common/php/screen-options.php#L169

I think this:
`$id = end(explode('-', $_POST['action'], 2));`

Should be split into two lines:

```
$ids = explode('-', $_POST['action'], 2);
$id = end($ids);
```

PR #667, opened 2021-06-10
Fixes non-variable being passed by reference

mikeyarce:
## Description

Fixes https://github.com/Automattic/Edit-Flow/issues/656

Currently there's a PHP Notice happening when you save screen option changes:
`[07-Jun-2021 18:36:16 UTC] PHP Notice:  Only variables should be passed by reference in wp-content/plugins/Edit-Flow/common/php/screen-options.php on line 169`

## Steps to Test

1. Go to `wp-admin/index.php?page=story-budget`
2. Click on Screen Options at the top
3. Change the number of columns
4. Refresh the page to make sure the setting was saved

### DIFF
--- a/common/php/screen-options.php
+++ b/common/php/screen-options.php
@@ -166,7 +166,8 @@ class wsScreenOptions10 {
 		}
 		
 		//The 'action' argument is in the form "save_settings-panel_id"
-		$id = end(explode('-', $_POST['action'], 2));
+		$ids = explode( '-', $_POST['action'], 2 );
+		$id = end( $ids );
 		
 		//Basic security check.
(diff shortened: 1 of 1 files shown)


mikeyarce:
Looking at the test failures, it looks like the problem is:
`Line indented incorrectly; expected at least 3 tabs, found 2`

The problem happens because the whole file is indented incorrectly, but I don't think we should worry about that for this specific PR.

htdat:
Tested and confirmed everything is OK. 
🚢 !

htdat (approved):
🚢! 

(I noticed that I did not formally send this review so... I am doing that now).

Issue #662, opened 2021-06-09
jQuery - redundant code in modules/custom-status/lib/custom-status.js


htdat:
From https://github.com/Automattic/Edit-Flow/pull/649/files#r645296481

---

## Expected/Desired Behavior

Remove the redundancy code. 

## Code: 

http://github.com/Automattic/Edit-Flow/blob/8a52203127b483576410558f940097e583d6ef3d/modules/custom-status/lib/custom-status.js#L84-L87

## Details 

**tl;dr: We can remove this jQuery listener because (1) jQuery selector is no longer found, and (2) there is no issue with custom statuses and `Quick Edit` right now.** 

As mentioned in the comment here, this `click` event is triggered if `Quick Edit` links are clicked in http://valet-vipgo-3279.test/wp-admin/edit.php

However, `Quick Edit` is currently a button element. Its HTML looks like this: 

```html
<button type="button" class="button-link editinline" aria-label="Quick edit “Add a new post – set this for future” inline" aria-expanded="false">Quick&nbsp;Edit</button>
```
Also, in the browser console, I can double check that with these functions:

- `jQuery('#the-list a.editinline').length` returns 0
- `jQuery('#the-list button.editinline').length` returns `20` (or the number of posts on this screen).

That said, I am not seeing any issue with `Quick Edit` and custom statuses here.

PR #668, opened 2021-06-11
Remove redundant jQuery selector #the-list a.editinline

htdat:
Fix https://github.com/Automattic/Edit-Flow/issues/662

---

## Description

This is pretty straightforward as all research has been done in https://github.com/Automattic/Edit-Flow/issues/662

## Steps to Test

1. Apply this PR.
2. Apply the diff in the **Note** section for `edit_flow.php` to avoid caching issue for JS files https://github.com/Automattic/Edit-Flow/pull/649
3. Visit wp-admin > Posts

✅ if we can use `Quick Edit` and change custom status without any issue.

### DIFF
--- a/modules/custom-status/lib/custom-status.js
+++ b/modules/custom-status/lib/custom-status.js
@@ -81,10 +81,6 @@ jQuery(document).ready(function() {
 
 	} else if ( jQuery('select[name="_status"]').length > 0 ) {
 		ef_append_to_dropdown('select[name="_status"]');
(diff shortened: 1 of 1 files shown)


mikeyarce (approved):
Thanks for this, @htdat ! 

I tested this out and Quick Edit works as expected.   I'm curious as to why this was needed originally...  My guess is it had something to do with what WordPress did maybe 10 years ago but is no longer the case.

Either way - Quick Edit custom statuses are loaded correctly and I'm able to select them, save them, change them, etc. 
👍🏽

PR #674, opened 2021-06-24
bring actions setup in line with other plugins

trepmal:
Aims to bring some GitHub actions setup inline with other Automattic & VIP plugins


GaryJones (approved):

.github/workflows/cs-lint.yml
```diff
@@ -0,0 +1,70 @@
...
+      - name: Lint phpunit.xml.dist
+        uses: ChristophWurst/xmllint-action@v1
+        with:
+          xml-file: ./phpunit.xml.dist
+          xml-schema-file: ./vendor/phpunit/phpunit/phpunit.xsd
+
+      # Check the code-style consistency of the PHP files.
+#      - name: Check PHP code style
+#        continue-on-error: true
+#        run: vendor/bin/phpcs --report-full --report-checkstyle=./phpcs-report.xml
```
htdat, line 67: @trepmal
This is how Edit Flow is implementing to check PHPCS for just diff files. 

https://github.com/Automattic/Edit-Flow/blob/c7e3fa7536b5787d494d54d3e34f240355e06dc6/.github/workflows/php-tests.yml#L56-L60

Ideally, we run PHPCS for all PHP files, but for old plugins like this, it's not productive to detect and fix all PHPCS errors. So I think this approach should be still kept though there may be a better way.

GaryJones:
I'm working on bringing consistency across the VIP supported plugins, so as there are already conflicts, I'll close this out.

PR #721, opened 2024-05-17
Add the ability to expire content on a date

ingeniumed:
## Description

This PR is meant to add the ability to expire the content of a post, on a set date. This is done by:

- Introducing a new custom status called expired
- Introducing a new field called expiry date

When the expiry date is set, a task is made to set the post's status as expired. That in turn, if notifications have been subscribed to, will notify groups or users.

The intention is that, we want to slowly build in more workflow features on top of this plugin and this is our stepping stone to that.

## Steps to Test

1. Draft up a new post.
2. Set the expiry date on it.
3. Publish it.
4. Check the cron list using wp-cli to ensure a task with the name `ef_unpublish_post_task` has been made.
5. Go in and update the expiry date on that post.
6. Check the cron list using wp-cli to ensure the task's date has been updated.
7. Wait for the task's date to pass.
8. Check the post to see that it's status was changes to `expired`. This will be visible in the story budget view.
9. Repeat this test, but this time delete the expired status and ensure the status is set to `pending`.

Note: If you are using the `vip dev-env` to test this PR, you'll need to manually run the task as cron doesn't run with the dev-env.

### DIFF
--- a/modules/custom-status/custom-status.php
+++ b/modules/custom-status/custom-status.php
@@ -166,7 +166,7 @@ class EF_Custom_Status extends EF_Module {
 				'term' => __( 'Expired', 'edit-flow' ),
 				'args' => array(
 					'slug' => 'expired',
-					'description' => __( 'Post has expired, and it has to bew updated an editor.', 'edit-flow' ),
+					'description' => __( 'Post has expired and requires updates .', 'edit-flow' ),
 					'position' => 6,
 				),
 			),
--- a/modules/custom-status/custom-status.php
+++ b/modules/custom-status/custom-status.php
@@ -162,6 +162,14 @@ class EF_Custom_Status extends EF_Module {
 					'position' => 5,
 				),
 			),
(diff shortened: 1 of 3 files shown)


modules/custom-status/custom-status.php
```diff
@@ -162,6 +162,14 @@ function install() {
 					'position' => 5,
 				),
 			),
+			array(
+				'term' => __( 'Expired', 'edit-flow' ),
+				'args' => array(
+					'slug' => 'expired',
+					'description' => __( 'Post has expired, and it has to bew updated an editor.', 'edit-flow' ),
```
alecgeatches, line 169: Think there's a typo. Here's a possible condensed version:

```suggestion
					'description' => __( 'Post has expired and requires updates .', 'edit-flow' ),
```

ingeniumed:
Closing this off for now as we will work on this again in the future.

PR #723, opened 2024-05-21
Fix the failing linting step

ingeniumed:
## Description

As the next step for modernization, this PR:

- Updates eslint packages, and adds in the rules from `@automattic/eslint-plugin-wpvip` as a starting point. There are a lot of rules being switched off in order to get the linting to pass. This will slowly be fixed as the underlying code is updated as well as the node/installed packages.
- Runs linting on all the JS files and updates the build as a result
- Removes the wordpress prettier packages as the above eslint plugin has that built in.


ingeniumed:
Note: The failing test is a flaky test that passes when it's re-run

alecgeatches:
@ingeniumed This looks good! I noticed some libraryish files in `common/js` have been formatted (probably through `--fix`), but it'd probably make more sense to revert and ignore those.

I'm looking into that flaky test failure now.

.eslintrc.js
```diff
@@ -0,0 +1,38 @@
+require( '@automattic/eslint-plugin-wpvip/init' );
+
+module.exports = {
+	extends: [ 'plugin:@automattic/wpvip/recommended' ],
+	root: true,
+	env: {
+		jest: true,
+	},
+	rules: {
+		"no-prototype-builtins": 0,
```
alecgeatches, line 10: @ingeniumed Do you think any of these rules are more important than the others and we can apply them now? Or maybe we can make issues to address the more important exceptions (`no-eval`) soon. I get ignoring huge refactorings like `valid-jsdoc`, but it'd be great if low-hanging code quality and safety rules would be applied.

ingeniumed:
> @ingeniumed This looks good! I noticed some libraryish files in `common/js` have been formatted (probably through `--fix`), but it'd probably make more sense to revert and ignore those entirely.
> 
> I'm looking into that flaky test failure now.

Good point, I'll revert that. In another PR it might be worth seeing if there's a built in equivalent of that library now, as that could allow us to drop that code.

.eslintrc.js
ingeniumed: I wanted to avoid adding more scope to this PR by resolving those, that's really why I avoided adding it in here. I also think as we build more features into this plugin, we are likely to change/upgrade that code (which might happen anyways thanks to the package modernizations)

alecgeatches (approved):
👍

PR #726, opened 2024-05-24
Bump plugin version to 0.9.9

alecgeatches:
Bumps plugin version to `0.9.9` based off of these recent PR merges into `develop`:

* Test fix: Fix e2e tests, bump lowest supported PHP version to 7.4 (https://github.com/Automattic/Edit-Flow/pull/722)
* Test fix: Update ESLint configuration and format JS files (https://github.com/Automattic/Edit-Flow/pull/723)
* Enhancements: Update WordPress minimum version to 5.9 & update PHPUnit #724 (https://github.com/Automattic/Edit-Flow/pull/724)
* Enhancements: Move JS environment to node 20, upgrade packages (https://github.com/Automattic/Edit-Flow/pull/725)
* README updates from this branch: https://github.com/Automattic/Edit-Flow/pull/726/commits/579a365ad0973fd1074f72909221bb3d8615eb19

PR #727, opened 2024-05-28
Bump the minimim version of php to 8.0

ingeniumed:
## Description

Bump the minimum version of php to 8.0, as well as the minimum WP version to 6.0. As a result of this, the packages in `composer.json` have been updated and the GH workflow has been modified to run against 8.0/6.0 rather than 7.4/5.9.

### DIFF
--- a/edit_flow.php
+++ b/edit_flow.php
@@ -1,31 +1,17 @@
 <?php
-/*
-Plugin Name: Edit Flow
-Plugin URI: http://editflow.org/
-Description: Remixing the WordPress admin for better editorial workflow options.
-Author: Daniel Bachhuber, Scott Bressler, Mohammad Jangda, Automattic, and others
-Version: 0.9.8
-Author URI: http://editflow.org/
-
-Copyright 2009-2019 Mohammad Jangda, Daniel Bachhuber, Automattic, et al.
-
-GNU General Public License, Free Software Foundation <http://creativecommons.org/licenses/GPL/2.0/>
-
-This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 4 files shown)


alecgeatches (approved):
Looks good! We can address the failing e2e tests (unrelated) in https://github.com/Automattic/Edit-Flow/pull/726.

edit_flow.php
```diff
@@ -1,31 +1,17 @@
...
+/**
+ * Plugin Name: Edit Flow
+ * Plugin URI: http://editflow.org/
+ * Description: Remixing the WordPress admin for better editorial workflow options.
+ * Author: Daniel Bachhuber, Scott Bressler, Mohammad Jangda, Automattic, and others
+ * Version: 0.9.8
+ * Requires at least: 6.0
+ * Requires PHP: 8.0
+ * License: GPL-3
+ * License URI: https://www.gnu.org/licenses/gpl-3.0.html
```
swissspidy, line 11: Was the license change intentional?

`composer.json` still says GPL v2 or later. What is it?

I'd recommend adding a proper LICENSE file to the repo.

ingeniumed: Thanks for catching that. We meant to make it GPL 3 as some of our other plugins have that already, so we were trying to make it consistent. I'll fix that inconsistency, and add a proper LICENSE file to the repo.

GaryJones: This was an incorrect decision. A site that uses a GPLv2-only plugin can't legally be allowed to work with a version of Edit-Flow that is GPLv3 only. I've opened #804 to rectify this.

PR #730, opened 2024-06-11
Add support for Slack notifications

hanifn:
## Description

This PR adds support for Slack notifications on top of EditFlow's existing email notifications. Currently we're only planning to support Slack's incoming webhook API.

## Steps to Test

You would need to create a Slack app on your own workspace and enable Incoming Webhooks for that app.

1. Check out PR.
1. Go to Notifications.
1. Enable `Send to Webhook` and add in the webhook URL.
1. Save settings
1. Change a post status or add editorial comments.
1. Verify that Slack messages were sent to the right channel on your Slack workspace


modules/notifications/notifications.php
```diff
@@ -678,35 +680,38 @@ function notification_status_change( $new_status, $old_status, $post ) {
...
 			$body .= sprintf( __( 'Edit: %s', 'edit-flow' ), $edit_link ) . "\r\n";
 			$body .= sprintf( __( 'View: %s', 'edit-flow' ), $view_link ) . "\r\n";
-				
+
 			$body .= $this->get_notification_footer($post);
-			
+
 			$this->send_email( 'status-change', $post, $subject, $body );
+
+			if ( 'on' === $this->module->options->send_to_slack )
```
alecgeatches: @hanifn Minor, can you change any new `if`s to use braces? I see the plugin is doing braceless `if`s all over the place but I'd prefer any new code avoid that.

```suggestion
			if ( 'on' === $this->module->options->send_to_slack ) {
			    $this->send_to_slack( $body );
			}
```

modules/notifications/notifications.php
```diff
@@ -740,35 +745,41 @@ function notification_comment( $comment ) {
...
+		$body .= "\r\n" . sprintf( __( 'You can see all editorial comments on this %s here: ', 'edit-flow' ), $post_type ). "\r\n";
 		$body .= $edit_link . "#editorialcomments" . "\r\n\r\n";
-		
+
 		$body .= $this->get_notification_footer($post);
-		
+
 		$this->send_email( 'comment', $post, $subject, $body );
+
+		$format = "*%s* left a comment on *%s #%s - %s*\n\n %s";
```
alecgeatches: We could add a handful of i18n changes here. Can we:

1. Update this to a format string ( `__( '...', 'edit-flow' )` )
2. Use parameter numbers for interpolation (`%1$s`)
3. Add a `translators:` string?

Here's an example from another place in the plugin doing all three:

https://github.com/Automattic/Edit-Flow/blob/b7d70af548a9aff8eeca24e1d0e0d5121e1b8d08/modules/notifications/notifications.php#L652-L653

If you're new to WordPress i18n functions, [check out this page too](https://codex.wordpress.org/I18n_for_WordPress_Developers#Placeholders). Thanks!

modules/notifications/notifications.php
```diff
@@ -803,10 +814,33 @@ function send_email( $action, $post, $subject, $message, $message_headers = '' )
...
+	function send_to_slack( $message ) {
+		$slack_webhook_url = $this->module->options->slack_webhook_url;
+
+		// Bail if the webhook URL is not set
+		if ( empty( $slack_webhook_url ) ) {
+			return;
+		}
+
+		// Set up the payload
+		$payload = array(
```
alecgeatches: Suuuper minor nit, but can we use array literal notation `[`/`]` in new code instead of the `array()` syntax?

modules/notifications/notifications.php
```diff
@@ -1223,25 +1259,53 @@ function settings_post_types_option() {
...
+	 * Option to enable sending notifications to Slack
+	 *
+	 * @since 0.9.9
+	 */
+	function settings_send_to_slack() {
+		$options = array(
+			'off' => __( 'Disabled', 'edit-flow' ),
+			'on' => __( 'Enabled', 'edit-flow' ),
+		);
+		echo '<select id="send_to_slack" name="' . $this->module->options_group_name . '[send_to_slack]">';
```
alecgeatches:
```suggestion
		echo '<select id="send_to_slack" name="' . esc_attr( $this->module->options_group_name ) . '[send_to_slack]">';
```

modules/notifications/notifications.php
```diff
@@ -1223,25 +1259,53 @@ function settings_post_types_option() {
...
 		echo '</select>';
 	}
 
+	/**
+	 * Option to set the Slack webhook URL
+	 *
+	 * @since 0.9.9
+	 */
+	function settings_slack_webhook_url() {
+		echo '<input type="text" id="slack_webhook_url" name="' . $this->module->options_group_name . '[slack_webhook_url]" value="' . esc_attr( $this->module->options->slack_webhook_url ) . '" />';
```
alecgeatches:
```suggestion
		echo '<input type="text" id="slack_webhook_url" name="' . esc_attr( $this->module->options_group_name ) . '[slack_webhook_url]" value="' . esc_attr( $this->module->options->slack_webhook_url ) . '" />';
```

modules/notifications/notifications.php
hanifn: I actually prefer braces myself as it's clearer to read but I do want to keep the code consistent in style. I'm not sure if changing that is a good idea if we're not planning to change existing code as well. Perhaps it's time to write a style guide for this repo? 🤔

modules/notifications/notifications.php
hanifn: Yes I was planning to add i18n functions as well it's just that these text is not finalised yet. If you've noticed, that text isn't even in use 😄

modules/notifications/notifications.php
hanifn: As I've mentioned in another comment, I've opted to stick to current conventions in existing code to make it more consistent.

modules/notifications/notifications.php
hanifn: Changed in 8bf4f9f190d771d89f4cdaa2855b5b37929f9f63

modules/notifications/notifications.php
hanifn: Changed in 8bf4f9f190d771d89f4cdaa2855b5b37929f9f63

modules/notifications/notifications.php
alecgeatches: I get keeping consistency with a codebase's style, but I think this may be a good time to start changing it.

A style guide would be great, as well as upgrading our PHPCS rules. @ingeniumed and I had a discussion about this [in JS linting rules](https://github.com/Automattic/Edit-Flow/pull/723/files#r1608821536), where we decided we would avoid massive rewrites to meet modern linting rules, but instead incrementally improve code as we encounter it. I think we should do the same here. We can waterfall from the top to the bottom of the list depending on what makes sense:

1. Fix a problem in code for the whole plugin, or if that's out of scope
2. Fix the problem for the whole file, or at least
3. Improve the code for changed functions, or if that's too risky
4. Ensure new code follows modern standards

I think we should encode what is "a problem in code" with PHPCS rules and linting, but for now I think conditionals without braces [are a Bad Practice](https://www.codecentric.de/wissens-hub/blog/curly-braces) enough to ensure we don't write new code this way. Same with `array()` versus `[]` literals, there's no great reason to adopt 2010 PHP coding standards into new features. We may never do a rewrite of some older features to be safe, but we can at least ensure new code is up to date.

modules/notifications/notifications.php
alecgeatches: True! 😄

modules/notifications/notifications.php
ingeniumed: Agreed on the above. We should approach each PR with the goal of modernizing every file we touch, which would mean:

- Lint the code if it's in PHP
- Modernize the PHP code
- Remove the JQuery pieces

modules/notifications/notifications.php
ingeniumed: Def will echo what Alec said - our goal should not be consistency, but modernizing the code that we touch to the new 2024 standards. It's going to mean discrepancy between files until we finish our work, but the end result will be a slow march to modernization.

modules/notifications/notifications.php
```diff
@@ -10,21 +10,21 @@
 if ( !class_exists('EF_Notifications') ) {
 
 class EF_Notifications extends EF_Module {
-	
+
```
ingeniumed: Would lint this class. If you have the save on format setup in VSCode, you should be able to hit save and have everything be formatted

modules/notifications/notifications.php
```diff
@@ -10,21 +10,21 @@
...
 	// Taxonomy name used to store users following posts
 	var $following_users_taxonomy = 'following_users';
```
ingeniumed: Modernization example: would be a good time to drop the use of var, and replace it with it's modern equivalent of private, protected or public

modules/notifications/notifications.php
hanifn: Understood 👍 
I also agree with what @alecgeatches said and if we're already planning to overhaul everything that I think this is a fine way to start.

modules/notifications/notifications.php
hanifn: Yeah I think my save on format setup in VSCode isn't working properly. It seems to be applying the prettier config I've set up for our NodeJS projects instead.

modules/notifications/notifications.php
hanifn: Hmm looks like our phpcs settings are not allowing the array literal notation: `Short array syntax is not allowed`

hanifn:
So it turns out I had to refactor the entire file anyway because my initial changes triggered some indentation errors in the phpcs diff checks. Simply fixing the indentation on a few select lines obviously wouldn't work so I had to fix the indent of the entire class which then increased the size of the diff so I had to fix everything. Now the only things remaining are some  `Processing form data without nonce verification.` errors which I'm unsure how to fix. Either because the nonce check was already done earlier in the method or the endpoints weren't receiving a nonce in the first place. 🤔

blocks/dist/custom-status.build.js
ingeniumed, line 1: Did you run `npm run build` to generate these? I ran the same locally, and it reverts back to the way these were before. This could be the dev build, instead of the production build.

modules/notifications/lib/notifications.js
```diff
@@ -87,4 +87,17 @@ jQuery( document ).ready( function ( $ ) {
 			} );
 		}
 	);
+
```
ingeniumed, line 90: I'm thinking that we should start adding ToDos here, saying to convert this to non-JQuery code. That way, we can have some entry points on hand for how to do the migration away from JQuery.

modules/notifications/notifications.php
```diff
@@ -4,1292 +4,1453 @@
...
+		public function get_user_following_posts( $user = 0, $args = null ) {
+			if ( ! $user ) {
+				$user = (int) wp_get_current_user()->ID;
+			}
+
+			if ( is_int( $user ) ) {
+				$user = get_userdata( $user )->user_login;
+			}
+
+			$post_args = array(
```
ingeniumed: Hitting save on VSCode for me changes the formatting slightly for this block of code. Running PHPCS doesn't bring up anything though, so I think this might be prettify doing that?

modules/notifications/notifications.php
```diff
@@ -4,1292 +4,1453 @@
...
+		 * Register the module with Edit Flow but don't do anything else
+		 */
+		public function __construct() {
+
+			// Register the module with Edit Flow
+			$this->module_url = $this->get_module_url( __FILE__ );
+			$args             = array(
+				'title'                 => __( 'Notifications', 'edit-flow' ),
+				'short_description'     => __( 'Update your team of important changes to your content.', 'edit-flow' ),
+				'extended_description'  => __( 'With email notifications, you can keep everyone updated about what’s happening with a given content. Each status change or editorial comment sends out an email notification to users subscribed to a post. User groups can be used to manage who receives notifications on what. With webhook notifications, all notifications will also be sent to the specified webhook URL(i.e.: Slack incoming webhooks).', 'edit-flow' ),
```
ingeniumed: This description should be tweaked a touch to reflect that notifications is the purpose of this module and that's done in two ways:

- Email to a group or a batch of users
- Webhook which isn't piped to a specific group or a batch

Right now that's slightly confusing, since webhook flow doesn't support users/groups.

modules/notifications/notifications.php
```diff
@@ -4,1292 +4,1453 @@
...
-		if ( apply_filters( 'ef_notifications_show_follow_link', true ) ) {
-			// A little extra JS for the follow button
-			add_action( 'admin_head', array( $this, 'action_admin_head_follow_js' ) );
-			// Manage Posts
-			add_filter( 'post_row_actions', array( $this, 'filter_post_row_actions' ), 10, 2 );
-			add_filter( 'page_row_actions', array( $this, 'filter_post_row_actions' ), 10, 2 );
-			// Calendar and Story Budget
-			add_filter( 'ef_calendar_item_actions', array( $this, 'filter_post_row_actions' ), 10, 2 );
-			add_filter( 'ef_story_budget_item_actions', array( $this, 'filter_post_row_actions' ), 10, 2 );
+				'settings_help_sidebar' => __( '<p><strong>For more information:</strong></p><p><a href="http://editflow.org/features/notifications/">Notifications Documentation</a></p><p><a href="http://wordpress.org/tags/edit-flow?forum_id=10">Edit Flow Forum</a></p><p><a href="https://github.com/danielbachhuber/Edit-Flow">Edit Flow on Github</a></p>', 'edit-flow' ),
```
ingeniumed: Minor: We should change the danielbachhuber to automattic here.

modules/notifications/notifications.php
```diff
@@ -4,1292 +4,1453 @@
...
+
+			// Send the notification
+			$response = wp_remote_post(
+				$webhook_url,
+				array(
+					'body'    => wp_json_encode( $payload ),
+					'headers' => array( 'Content-Type' => 'application/json' ),
+				)
+			);
+			if ( is_wp_error( $response ) ) {
```
ingeniumed, line 908: I tried to test the error pathway by giving an invalid webhook url, and the commenting system just hangs as a result. It seems like it's not able to parse the ajax response, and as a result it just gets stuck on an endless spinner. This seems like it's happening in the editorial-comments.js.

ingeniumed: In addition, should we print the ajax response? Wondering if that would reveal some sensitive info, and instead if we should just say that "Sending to webhook failed" and instead send some extra debugging info to the error logs

ingeniumed (commented):
None of the below points have to be done in this PR, they are for follow up PRs.

Some notes I noted while testing the UI:

- The UI is tuned towards sending email notifications to a user or group. If none of those are to checked then there is a default message of "no one will be notified" that's to be displayed. In the follow up PRs this is something we should address if send to webhook is enabled.
- For cleaning up the email UI, I'm thinking we introduce a checkbox that says email enabled whose default value is true. If you want, you can flip it to false and that will hide that meta box entirely.

We can make the webhook functionality a bit more robust with the following:

- We should have some sort of disabling functionality for the webhook, in the event that it fails more than once. That way, we don't need to retry it and they'd need to go in and fix the URL or the problem instead.
- We should also introduce a retry functionality
- Right now it's a synchronous call if I'm correct. Could we consider switching to an async call so it's not blocking?

blocks/dist/custom-status.build.js
ingeniumed: I ended up running it and pushing it up btw. So it's fixed now

modules/notifications/notifications.php
hanifn: I've fixed the error handling to not hang the UI in 57a6806a6e8761cf71e06404c7d05c2728f86731 but the error display is kinda terrible. I've tried to style it as best as I can but I'm still no happy with it.

Screenshot 2024-06-19 at 2 59 10 PM

ingeniumed: What if we ran a check when the webhook was saved, so we could show the error message then and there? That way, it's easy to prevent a save if the webhook is invalid and we could change this to something like "Unable to send notification to webhook provided"

modules/notifications/notifications.php
ingeniumed: You might need to update that to be similar to the governance plugin [here](https://github.com/Automattic/vip-governance-plugin/blob/trunk/.phpcs.xml.dist#L62)

hanifn: Are you saying we should override the existing WordPress-VIP-Go ruleset to support short array syntax?

modules/notifications/notifications.php
hanifn: I'm not sure what you mean by that @ingeniumed, can you give an example?

modules/notifications/notifications.php
ingeniumed: Yeah, that's what I'm saying. Adding that exclusion rule that I linked will allow you to use the short array syntax

modules/notifications/notifications.php
ingeniumed: So what I'm suggesting is two fold:

- When the slack webhook url is provided, and the save button is pressed we run a test on that url. We post a test message that says "Integrated successfully with EditFlow". If this doesn't happen, the save fails then and there. This will reduce the chances of failures happening down the line (doesn't eliminate it but sure helps).
- When a notification fails to be sent in the block editor,  we change the error that's shown to be more specific i.e. "Unable to send notification to webhook provided".

Hope that clears up what I mean.

modules/notifications/notifications.php
hanifn: Added in 34906df60e78b9d901c3c1c2ee9e94ede312ec4a

modules/notifications/lib/notifications.js
hanifn: Added in 6458d6b18d0a0ca671d868e82e04626b2b921711

modules/notifications/notifications.php
hanifn: Changed in 69676fd4e45c74b8609102c2820c9136bce8d350

modules/notifications/notifications.php
hanifn: Changed in 69676fd4e45c74b8609102c2820c9136bce8d350

modules/notifications/notifications.php
hanifn: Gotcha. I get it now 👍

modules/notifications/notifications.php
```diff
@@ -4,1292 +4,1509 @@
...
+		 *
+		 * @todo add_cap to set subscribers for posts; default to Admin and editors
+		 */
+		public function notifications_meta_box() {
+			global $post, $post_ID, $edit_flow;
+			?>
+			<div id="ef-post_following_box">
+				<a name="subscriptions"></a>
+
+				<p><?php _e( 'Select the users and user groups that should receive notifications when the status of this post is updated or when an editorial comment is added.', 'edit-flow' ); ?></p>
```
ingeniumed: We should add email notifications here so it's clear that webhook doesn't count.

hanifn: Changed in 902bc0473ba0f728ba2b6f95abbf208be1e71650

ingeniumed (approved):
🚢

PR #731, opened 2024-06-21
Move the block editor JS/CSS blocks for the custom status

ingeniumed:
## Description

This will make the block editor version of the custom status module the default one. It uses the newer actions to load it, in just the block editor. In addition, it also moves the files to the module and loads everything that's compiled in the generic dist folder. The intention is that once we convert everything away from JQuery, it can just sit in that dist folder.

I've also disabled the PHPCSDiff script just in the interest of making it easier for us to update each of the modules. We can re-enable this later, once we have re-done each of the modules.

## Steps to Test

- Re-test the custom status module, it should work just like it did before


modules/custom-status/custom-status.php
```diff
@@ -321,7 +315,30 @@ function action_admin_enqueue_scripts() {
 			));
 		}
 
+	}
+
+	function load_scripts_for_block_editor(){
+		// ToDo: This is duplicated in 3 places, should be refactored
```
hanifn: Maybe we can move this check into the `init()` method above? Then it wouldn't be duplicated so much. Something like:

```
if ( ! $this->disable_custom_statuses_for_post_type() ) {
	// Load CSS and JS resources that we probably need in the admin page
	add_action( 'admin_enqueue_scripts', array( $this, 'action_admin_enqueue_scripts' ) );

	// Assets for block editor UI.
	add_action( 'enqueue_block_editor_assets', array( $this, 'load_scripts_for_block_editor') );

	// Assets for iframed block editor and editor UI.
	add_action( 'enqueue_block_editor_assets', array( $this, 'load_styles_for_block_editor') );
}
```

hanifn (approved):
LGTM 👍 Custom status is still working as expected.
Just have a non-blocking suggestion to your duplicated code question

modules/custom-status/custom-status.php
ingeniumed: Good call, makes sense to me

ingeniumed:
I have re-enabled the php cs diff test but I haven't fixed them in this PR because it's over 2000 lines of changes. I'll follow up with another PR that focuses on that, so it's easy to revert any build problems in the JS vs the php side.

PR #732, opened 2024-06-24
Add the vip feature flag, analytics flag and bump plugin version

ingeniumed:
## Description

- Add the vip feature flag, as well as the analytics flag. These don't do anything right now, but they have helper methods tied to them at the module level so it's easy to see if its enabled or not. In addition, this is set to on by default if its a VIP site. It's shown on the settings page itself.
- Bump up the plugin version to 0.10.0 so that we can properly mark the since fields.
- Remove `timesince` because it's not used and it's not really do proper GMT calculations. The old code that it used to be used for has been replaced by [human_time](https://developer.wordpress.org/reference/functions/human_time_diff/) which is way simpler. Didn't really think marking this as deprecated and pointing to that is better. We will encounter more cases like this and we can make a call case by case.
- Lots of PHPCS fixes related to escaping the output, and disabling the nonce for the settings controller rendering.

## Steps to Test

- Play around with the settings for vip_features and analytics to make sure it's okay.
- Everything else should be unaffected


common/php/class-module.php
```diff
@@ -5,296 +5,338 @@
...
+		/**
+		 * Check if the site is a WPVIP site.
+		 *
+		 * @since 0.10.0
+		 * @return true, if it is a WPVIP site, false otherwise
+		 */
+		protected function is_vip_site() {
+			return defined( 'WPCOM_IS_VIP_ENV' ) && constant( 'WPCOM_IS_VIP_ENV' ) === true
+			&& defined( 'WPCOM_SANDBOXED' ) && constant( 'WPCOM_SANDBOXED' ) === false
+			&& defined( 'FILES_CLIENT_SITE_ID' );
```
alecgeatches: @ingeniumed I noticed in `vip dev-env` that `WPCOM_IS_VIP_ENV` is `false`, it seems like it's only set to `true` on production sites. Can we change this to use the presence of `VIP_GO_ENV` instead? The valid values for this seem to be `"production"` and `"local"`, but I think just checking that it exists will do the trick:

```suggestion
			return defined( 'VIP_GO_ENV' )
			&& defined( 'WPCOM_SANDBOXED' ) && constant( 'WPCOM_SANDBOXED' ) === false
			&& defined( 'FILES_CLIENT_SITE_ID' );
```

Note that this is different than [the logic we used in the block data API](https://github.com/Automattic/vip-block-data-api/blob/05134af5f95a9fe055a205374943a2a842f35698/src/analytics/analytics.php#L99-L104), and that's mostly because we didn't want to send analytics locally. However, I think we should have the "Turn on WordPress VIP features" default to true on `vip dev-env`.

That probably means we'll want [two VIP-related checks](https://github.com/Automattic/Edit-Flow/blob/531140239dfdaa55d588ceb4834cd7f6f8cd10fd/modules/settings/settings.php#L25-L26), one for analytics (exclude local environments), and one for VIP features (local environments are fine).

alecgeatches (changes requested):
Left one small comment about VIP detection for `vip dev-env`, and I also see there are [some PHPCS errors](https://github.com/Automattic/Edit-Flow/actions/runs/9640875417/job/26585426657?pr=732) from the tests. Other than that, looks good to me! Glad we're getting a lot of the PHPCS formatting out of the way, because it's a lot to review and GitHub's diffing is not equipped to review this many file changes.

common/php/class-module.php
ingeniumed: Good catch, corrected this in [this commit](https://github.com/Automattic/Edit-Flow/pull/732/commits/60f04b0c947fcdcf95423049e0e7278963abf892) by allowing for a flag to be passed in that will ensure only production sites are considered wpvip sites.

ingeniumed:
> Left one small comment about VIP detection for `vip dev-env`, and I also see there are [some PHPCS errors](https://github.com/Automattic/Edit-Flow/actions/runs/9640875417/job/26585426657?pr=732) from the tests. Other than that, looks good to me! Glad we're getting a lot of the PHPCS formatting out of the way, because it's a lot to review and GitHub's diffing is not equipped to review this many file changes.

I ended up disabling that script for now. I figured once we lint the entire codebase we can renable it just so we don't keep getting failures. It makes sense it would fail though because linting custom status causes aboutg 1500 lines to change and that can be pushed off to another PR

PR #734, opened 2024-06-25
Lint the various modules

ingeniumed:
## Description

This PR is meant to lint all the PHP classes at once to make development easier. I have skipped the commons folder, just like we skipped it for the JS linting in the interest of not breaking everything. We should revisit that decision or try to refactor it, but I'll set that aside for now.

There's also a bug fix here as the settings updated css wasn't being applied correctly. That's resolved now.

### DIFF
--- a/modules/calendar/calendar.php
+++ b/modules/calendar/calendar.php
@@ -5,697 +5,706 @@
  *
  * @author danielbachhuber
  */
-if ( !class_exists('EF_Calendar') ) {
+if ( ! class_exists( 'EF_Calendar' ) ) {
 
-class EF_Calendar extends EF_Module {
+	class EF_Calendar extends EF_Module {
 
-	const usermeta_key_prefix = 'ef_calendar_';
-	const screen_id = 'dashboard_page_calendar';
+		// phpcs:ignore Generic.NamingConventions.UpperCaseConstantName.ClassConstantNotUpperCase
+		const usermeta_key_prefix = 'ef_calendar_';
+		// phpcs:ignore Generic.NamingConventions.UpperCaseConstantName.ClassConstantNotUpperCase
+		const screen_id = 'dashboard_page_calendar';
 
-	var $module;
+		public $module;
 
-	var $start_date = '';
-	var $current_week = 1;
-	var $total_weeks = 6; // default number of weeks to show per screen
-	var $hidden = 0; // counter of hidden posts per date square
-	var $max_visible_posts_per_date = 4; // total number of posts to be shown per square before 'more' link
+		public $start_date = '';
+		public $current_week = 1;
+		public $total_weeks = 6; // default number of weeks to show per screen
+		public $hidden = 0; // counter of hidden posts per date square
+		public $max_visible_posts_per_date = 4; // total number of posts to be shown per square before 'more' link
 
-	private $post_date_cache = array();
-	private static $post_li_html_cache_key = 'ef_calendar_post_li_html';
-	private int $max_weeks;
-	private string $create_post_cap;
+		private $post_date_cache = array();
+		private static $post_li_html_cache_key = 'ef_calendar_post_li_html';
+		private int $max_weeks;
+		private string $create_post_cap;
 
-	/**
-	 * Calendar published statuses are the same as other
-	 * components but without the future
-	 */
-	public $published_statuses = array(
-		'publish',
-		'private',
-	);
-
-	/**
-	 * Construct the EF_Calendar class
-	 */
-	function __construct() {
-		$this->max_weeks = 12;
-
-		$this->module_url = $this->get_module_url( __FILE__ );
-		// Register the module with Edit Flow
-		$args = array(
-			'title' => __( 'Calendar', 'edit-flow' ),
-			'short_description' => sprintf( __( 'View upcoming content in a <a href="%s">customizable calendar</a>.', 'edit-flow' ), admin_url( 'index.php?page=calendar' ) ),
-			'extended_description' => __( 'Edit Flow’s calendar lets you see your posts over a customizable date range. Filter by status or click on the post title to see its details. Drag and drop posts between days to change their publication date.', 'edit-flow' ),
-			'module_url' => $this->module_url,
-			'img_url' => $this->module_url . 'lib/calendar_s128.png',
-			'slug' => 'calendar',
-			'post_type_support' => 'ef_calendar',
-			'default_options' => array(
-				'enabled' => 'on',
-				'post_types' => array(
-					'post' => 'on',
-					'page' => 'off',
-				),
-				'quick_create_post_type' => 'post',
-				'ics_subscription' => 'off',
-				'ics_secret_key' => '',
-			),
-			'messages' => array(
-				'post-date-updated' => __( "Post date updated.", 'edit-flow' ),
-				'update-error' => __( 'There was an error updating the post. Please try again.', 'edit-flow' ),
-				'published-post-ajax' => __( "Updating the post date dynamically doesn't work for published content. Please <a href='%s'>edit the post</a>.", 'edit-flow' ),
-				'key-regenerated' => __( 'iCal secret key regenerated. Please inform all users they will need to resubscribe.', 'edit-flow' ),
-			),
-			'configure_page_cb' => 'print_configure_view',
-			'configure_link_text' => __( 'Calendar Options', 'edit-flow' ),
-			'settings_help_tab' => array(
-				'id' => 'ef-calendar-overview',
-				'title' => __('Overview', 'edit-flow'),
-				'content' => __('<p>The calendar is a convenient week-by-week or month-by-month view into your content. Quickly see which stories are on track to being published on time, and which will need extra effort.</p>', 'edit-flow'),
-				),
-			'settings_help_sidebar' => __( '<p><strong>For more information:</strong></p><p><a href="http://editflow.org/features/calendar/">Calendar Documentation</a></p><p><a href="http://wordpress.org/tags/edit-flow?forum_id=10">Edit Flow Forum</a></p><p><a href="https://github.com/danielbachhuber/Edit-Flow">Edit Flow on Github</a></p>', 'edit-flow' ),
-		);
-		$this->module = EditFlow()->register_module( 'calendar', $args );
-
-	}
-
-	/**
-	 * Initialize all of our methods and such. Only runs if the module is active
-	 *
-	 * @uses add_action()
-	 */
-	function init() {
-
-		// .ics calendar subscriptions
-		add_action( 'wp_ajax_ef_calendar_ics_subscription', array( $this, 'handle_ics_subscription' ) );
-		add_action( 'wp_ajax_nopriv_ef_calendar_ics_subscription', array( $this, 'handle_ics_subscription' ) );
-
-		// Check whether the user should have the ability to view the calendar
-		$view_calendar_cap = 'ef_view_calendar';
-		$view_calendar_cap = apply_filters( 'ef_view_calendar_cap', $view_calendar_cap );
-		if ( !current_user_can( $view_calendar_cap ) ) return false;
-
-		// Define the create-post capability
-		$this->create_post_cap = apply_filters( 'ef_calendar_create_post_cap', 'edit_posts' );
-
-		add_action( 'admin_init', array( $this, 'add_screen_options_panel' ) );
-		add_action( 'admin_init', array( $this, 'handle_save_screen_options' ) );
-
-		add_action( 'admin_init', array( $this, 'register_settings' ) );
-		add_action( 'admin_menu', array( $this, 'action_admin_menu' ) );
-		add_action( 'admin_print_styles', array( $this, 'add_admin_styles' ) );
-		add_action( 'admin_enqueue_scripts', array( $this, 'enqueue_admin_scripts' ) );
-
-		// Ajax manipulation for the calendar
-		add_action( 'wp_ajax_ef_calendar_drag_and_drop', array( $this, 'handle_ajax_drag_and_drop' ) );
-
-		// Ajax insert post placeholder for a specific date
-		add_action( 'wp_ajax_ef_insert_post', array( $this, 'handle_ajax_insert_post' ) );
-
-		//Update metadata
-		add_action( 'wp_ajax_ef_calendar_update_metadata', array( $this, 'handle_ajax_update_metadata' ) );
-
-		// Clear li cache for a post when post cache is cleared
-		add_action( 'clean_post_cache', array( $this, 'action_clean_li_html_cache' ) );
-
-		// Action to regenerate the calendar feed sekret
-		add_action( 'admin_init', array( $this, 'handle_regenerate_calendar_feed_secret' ) );
-
-		// Hacks to fix deficiencies in core
-		add_action( 'pre_post_update', array( $this, 'fix_post_date_on_update_part_one' ), 10, 2 );
-		add_action( 'post_updated', array( $this, 'fix_post_date_on_update_part_two' ), 10, 3 );
-	}
-
-	/**
-	 * Load the capabilities onto users the first time the module is run
-	 *
-	 * @since 0.7
-	 */
-	function install() {
-
-		// Add necessary capabilities to allow management of calendar
-		// view_calendar - administrator --> contributor
-		$calendar_roles = array(
-			'administrator' => array('ef_view_calendar'),
-			'editor' =>        array('ef_view_calendar'),
-			'author' =>        array('ef_view_calendar'),
-			'contributor' =>   array('ef_view_calendar')
+		/**
+		 * Calendar published statuses are the same as other
+		 * components but without the future
+		 */
+		public $published_statuses = array(
+			'publish',
+			'private',
 		);
 
-		foreach ( $calendar_roles as $role => $caps ) {
-			$this->add_caps_to_role( $role, $caps );
-		}
-	}
+		/**
+		 * Construct the EF_Calendar class
+		 */
+		public function __construct() {
+			$this->max_weeks = 12;
 
-	/**
-	 * Upgrade our data in case we need to
-	 *
-	 * @since 0.7
-	 */
-	function upgrade( $previous_version ) {
-		global $edit_flow;
-
-		// Upgrade path to v0.7
-		if ( version_compare( $previous_version, '0.7' , '<' ) ) {
-			// Migrate whether the calendar was enabled or not and clean up old option
-			if ( $enabled = get_option( 'edit_flow_calendar_enabled' ) )
-				$enabled = 'on';
-			else
-				$enabled = 'off';
-			$edit_flow->update_module_option( $this->module->name, 'enabled', $enabled );
-			delete_option( 'edit_flow_calendar_enabled' );
-
-			// Technically we've run this code before so we don't want to auto-install new data
-			$edit_flow->update_module_option( $this->module->name, 'loaded_once', true );
-		}
-
-	}
-
-	/**
-	 * Add the calendar link underneath the "Dashboard"
-	 *
-	 * @uses add_submenu_page
-	 */
-	function action_admin_menu() {
-		add_submenu_page('index.php', __('Calendar', 'edit-flow'), __('Calendar', 'edit-flow'), apply_filters( 'ef_view_calendar_cap', 'ef_view_calendar' ), $this->module->slug, array( $this, 'view_calendar' ) );
-	}
-
-	/**
-	 * Add any necessary CSS to the WordPress admin
-	 *
-	 * @uses wp_enqueue_style()
-	 */
-	function add_admin_styles() {
-		global $pagenow;
-		// Only load calendar styles on the calendar page
-		if ( 'index.php' === $pagenow && isset( $_GET['page'] ) && 'calendar' === $_GET['page'] ) {
-			wp_enqueue_style( 'edit-flow-calendar-css', $this->module_url . 'lib/calendar.css', false, EDIT_FLOW_VERSION );
-			wp_enqueue_style( 'edit-flow-calendar-react-css', $this->module_url . 'lib/dist/calendar.react.style.build.css', array( 'wp-components' ), EDIT_FLOW_VERSION );
-		}
-	}
-
-	/**
-	 * Add any necessary JS to the WordPress admin
-	 *
-	 * @since 0.7
-	 * @uses wp_enqueue_script()
-	 */
-	function enqueue_admin_scripts() {
-		global $pagenow;
-
-		if ( 'index.php' === $pagenow && isset( $_GET['page'] ) && 'calendar' === $_GET['page'] ) {
-			$this->enqueue_datepicker_resources();
-
-			$js_libraries = array(
-				'jquery',
-				'jquery-ui-core',
-				'jquery-ui-sortable',
-				'jquery-ui-draggable',
-				'jquery-ui-droppable',
-				'wp-data'
+			$this->module_url = $this->get_module_url( __FILE__ );
+			// Register the module with Edit Flow
+			$args = array(
+				'title' => __( 'Calendar', 'edit-flow' ),
+				/* translators: %s: URL to the calendar page */
+				'short_description' => sprintf( __( 'View upcoming content in a <a href="%s">customizable calendar</a>.', 'edit-flow' ), admin_url( 'index.php?page=calendar' ) ),
+				'extended_description' => __( 'Edit Flow’s calendar lets you see your posts over a customizable date range. Filter by status or click on the post title to see its details. Drag and drop posts between days to change their publication date.', 'edit-flow' ),
+				'module_url' => $this->module_url,
+				'img_url' => $this->module_url . 'lib/calendar_s128.png',
+				'slug' => 'calendar',
+				'post_type_support' => 'ef_calendar',
+				'default_options' => array(
+					'enabled' => 'on',
+					'post_types' => array(
+						'post' => 'on',
+						'page' => 'off',
+					),
+					'quick_create_post_type' => 'post',
+					'ics_subscription' => 'off',
+					'ics_secret_key' => '',
+				),
+				'messages' => array(
+					'post-date-updated' => __( 'Post date updated.', 'edit-flow' ),
+					'update-error' => __( 'There was an error updating the post. Please try again.', 'edit-flow' ),
+					/* translators: %s: URL to the published post */
+					'published-post-ajax' => __( "Updating the post date dynamically doesn't work for published content. Please <a href='%s'>edit the post</a>.", 'edit-flow' ),
+					'key-regenerated' => __( 'iCal secret key regenerated. Please inform all users they will need to resubscribe.', 'edit-flow' ),
+				),
+				'configure_page_cb' => 'print_configure_view',
+				'configure_link_text' => __( 'Calendar Options', 'edit-flow' ),
+				'settings_help_tab' => array(
+					'id' => 'ef-calendar-overview',
+					'title' => __( 'Overview', 'edit-flow' ),
+					'content' => __( '<p>The calendar is a convenient week-by-week or month-by-month view into your content. Quickly see which stories are on track to being published on time, and which will need extra effort.</p>', 'edit-flow' ),
+				),
+				'settings_help_sidebar' => __( '<p><strong>For more information:</strong></p><p><a href="http://editflow.org/features/calendar/">Calendar Documentation</a></p><p><a href="http://wordpress.org/tags/edit-flow?forum_id=10">Edit Flow Forum</a></p><p><a href="https://github.com/danielbachhuber/Edit-Flow">Edit Flow on Github</a></p>', 'edit-flow' ),
 			);
-			foreach( $js_libraries as $js_library ) {
-				wp_enqueue_script( $js_library );
+			$this->module = EditFlow()->register_module( 'calendar', $args );
+		}
+
+		/**
+		 * Initialize all of our methods and such. Only runs if the module is active
+		 *
+		 * @uses add_action()
+		 */
+		public function init() {
+
+			// .ics calendar subscriptions
+			add_action( 'wp_ajax_ef_calendar_ics_subscription', array( $this, 'handle_ics_subscription' ) );
+			add_action( 'wp_ajax_nopriv_ef_calendar_ics_subscription', array( $this, 'handle_ics_subscription' ) );
+
+			// Check whether the user should have the ability to view the calendar
+			$view_calendar_cap = 'ef_view_calendar';
+			$view_calendar_cap = apply_filters( 'ef_view_calendar_cap', $view_calendar_cap );
+			if ( ! current_user_can( $view_calendar_cap ) ) {
+				return false;
 			}
-			wp_enqueue_script( 'edit-flow-calendar-js', $this->module_url . 'lib/calendar.js', $js_libraries, EDIT_FLOW_VERSION, true );
 
-			$ef_cal_js_params = array( 'can_add_posts' => current_user_can( $this->create_post_cap ) ? 'true' : 'false' );
-			wp_localize_script( 'edit-flow-calendar-js', 'ef_calendar_params', $ef_cal_js_params );
+			// Define the create-post capability
+			$this->create_post_cap = apply_filters( 'ef_calendar_create_post_cap', 'edit_posts' );
 
-			/**
-			 * Powering the new React interface
-			 */
-			wp_enqueue_script( 'edit-flow-calendar-react-js', $this->module_url . 'lib/dist/calendar.react.build.js', array( 'react', 'react-dom', 'wp-components', 'wp-url', 'wp-data', 'moment' ), EDIT_FLOW_VERSION, true );
+			add_action( 'admin_init', array( $this, 'add_screen_options_panel' ) );
+			add_action( 'admin_init', array( $this, 'handle_save_screen_options' ) );
 
-			wp_add_inline_script(
-				'edit-flow-calendar-react-js',
-				'var EF_CALENDAR = ' . wp_json_encode( $this->get_calendar_frontend_config() ),
-				'before'
-			);
+			add_action( 'admin_init', array( $this, 'register_settings' ) );
+			add_action( 'admin_menu', array( $this, 'action_admin_menu' ) );
+			add_action( 'admin_print_styles', array( $this, 'add_admin_styles' ) );
+			add_action( 'admin_enqueue_scripts', array( $this, 'enqueue_admin_scripts' ) );
+
+			// Ajax manipulation for the calendar
+			add_action( 'wp_ajax_ef_calendar_drag_and_drop', array( $this, 'handle_ajax_drag_and_drop' ) );
+
+			// Ajax insert post placeholder for a specific date
+			add_action( 'wp_ajax_ef_insert_post', array( $this, 'handle_ajax_insert_post' ) );
+
+			//Update metadata
+			add_action( 'wp_ajax_ef_calendar_update_metadata', array( $this, 'handle_ajax_update_metadata' ) );
+
+			// Clear li cache for a post when post cache is cleared
+			add_action( 'clean_post_cache', array( $this, 'action_clean_li_html_cache' ) );
+
+			// Action to regenerate the calendar feed sekret
+			add_action( 'admin_init', array( $this, 'handle_regenerate_calendar_feed_secret' ) );
+
+			// Hacks to fix deficiencies in core
+			add_action( 'pre_post_update', array( $this, 'fix_post_date_on_update_part_one' ), 10, 2 );
+			add_action( 'post_updated', array( $this, 'fix_post_date_on_update_part_two' ), 10, 3 );
 		}
 
-	}
+		/**
+		 * Load the capabilities onto users the first time the module is run
+		 *
+		 * @since 0.7
+		 */
+		public function install() {
 
-	/**
-	 * Prepare the options that need to appear in Screen Options
-	 *
-	 * @since 0.7
-	 */
-	function generate_screen_options() {
+			// Add necessary capabilities to allow management of calendar
+			// view_calendar - administrator --> contributor
+			$calendar_roles = array(
+				'administrator' => array( 'ef_view_calendar' ),
+				'editor' => array( 'ef_view_calendar' ),
+				'author' => array( 'ef_view_calendar' ),
+				'contributor' => array( 'ef_view_calendar' ),
+			);
 
-		$output = '';
+			foreach ( $calendar_roles as $role => $caps ) {
+				$this->add_caps_to_role( $role, $caps );
+			}
+		}
 
-		$args = array(
+		/**
+		 * Upgrade our data in case we need to
+		 *
+		 * @since 0.7
+		 */
+		public function upgrade( $previous_version ) {
+			global $edit_flow;
+
+			// Upgrade path to v0.7
+			if ( version_compare( $previous_version, '0.7', '<' ) ) {
+				// Migrate whether the calendar was enabled or not and clean up old option
+				$enabled = get_option( 'edit_flow_calendar_enabled' );
+				if ( $enabled ) {
+					$enabled = 'on';
+				} else {
+					$enabled = 'off';
+				}
+				$edit_flow->update_module_option( $this->module->name, 'enabled', $enabled );
+				delete_option( 'edit_flow_calendar_enabled' );
+
+				// Technically we've run this code before so we don't want to auto-install new data
+				$edit_flow->update_module_option( $this->module->name, 'loaded_once', true );
+			}
+		}
+
+		/**
+		 * Add the calendar link underneath the "Dashboard"
+		 *
+		 * @uses add_submenu_page
+		 */
+		public function action_admin_menu() {
+			add_submenu_page( 'index.php', __( 'Calendar', 'edit-flow' ), __( 'Calendar', 'edit-flow' ), apply_filters( 'ef_view_calendar_cap', 'ef_view_calendar' ), $this->module->slug, array( $this, 'view_calendar' ) );
+		}
+
+		/**
+		 * Add any necessary CSS to the WordPress admin
+		 *
+		 * @uses wp_enqueue_style()
+		 */
+		public function add_admin_styles() {
+			global $pagenow;
+			// Only load calendar styles on the calendar page
+			if ( 'index.php' === $pagenow && isset( $_GET['page'] ) && 'calendar' === $_GET['page'] ) {
+				wp_enqueue_style( 'edit-flow-calendar-css', $this->module_url . 'lib/calendar.css', false, EDIT_FLOW_VERSION );
+				wp_enqueue_style( 'edit-flow-calendar-react-css', $this->module_url . 'lib/dist/calendar.react.style.build.css', array( 'wp-components' ), EDIT_FLOW_VERSION );
+			}
+		}
+
+		/**
+		 * Add any necessary JS to the WordPress admin
+		 *
+		 * @since 0.7
+		 * @uses wp_enqueue_script()
+		 */
+		public function enqueue_admin_scripts() {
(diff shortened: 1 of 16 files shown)


phpcs.xml.dist
```diff
@@ -17,16 +17,29 @@
...
 		<exclude name="WordPress.WP.CapitalPDangit.Misspelled" />
 		<!-- We use trigger_error extensively -->
 		<exclude name="WordPress.PHP.DevelopmentFunctions.error_log_trigger_error" />
+
+		<exclude name="WordPress.WP.I18n.InterpolatedVariableSingular" />
+		<exclude name="WordPress.WP.I18n.InterpolatedVariablePlural" />
+		<exclude name="WordPress.WP.I18n.MissingSingularPlaceholder" />
+		<exclude name="WordPress.WP.I18n.NonSingularStringLiteralText" />
+
+		<exclude name="Squiz.PHP.DisallowMultipleAssignments.FoundInControlStructure" />
```
hanifn, line 40: What is the rationale behind all these additional exclusions? Some of them, like `PEAR.NamingConventions.ValidClassName.StartWithCapital`, are pretty standard practise IMO. 
Maybe some explanatory comments will help here too?

ingeniumed: Good point, I just pushed up some comments for that so it's easier to understand. 

The tl;dr is that its for legacy reasons as we can't alter the problematic areas without breaking something. I'm hoping overtime we can find alternatives so this gives us a middle ground.

hanifn: > I'm hoping overtime we can find alternatives so this gives us a middle ground.

In that case, would it better to group all of these under a TODO for future reference?

ingeniumed: Good point, done in the latest commit

modules/calendar/calendar.php
```diff
@@ -749,111 +761,116 @@ function view_calendar() {
...
-						$week_dates[$i] = $week_single_date;
+					for ( $i = 0; $i < 7; $i++ ) {
+						$week_dates[ $i ] = $week_single_date;
 						$single_date_month = date_i18n( 'F', strtotime( $week_single_date ) );
 						if ( $single_date_month != $current_month ) {
 							$split_month = $single_date_month;
 							$current_month = $single_date_month;
 						}
-						$week_single_date = date( 'Y-m-d', strtotime( "+1 day", strtotime( $week_single_date ) ) );
+						$week_single_date = gmdate( 'Y-m-d', strtotime( '+1 day', strtotime( $week_single_date ) ) );
```
alecgeatches: @ingeniumed Changing from `date` to `gmdate` here leads to a date that's not in the timezone of the current site. For here (and other places) where `date` changed to `gmdate`, did we ensure behavior didn't change?

modules/calendar/calendar.php
```diff
@@ -952,51 +971,54 @@ function generate_post_li_html( $post, $post_date, $num = 0 ){
...
+				<?php foreach ( $this->get_post_information_fields( $post ) as $field => $values ) : ?>
 					<tr class="item-field item-information-<?php echo esc_attr( $field ); ?>">
 						<th class="label"><?php echo esc_html( $values['label'] ); ?>:</th>
-						<?php if ( $values['value'] && isset($values['type']) ): ?>
-							<?php if( isset( $values['editable'] ) && $this->current_user_can_modify_post( $post ) ) : ?>
-								<td class="value<?php if( $values['editable'] ) { ?> editable-value<?php } ?>"><?php echo esc_html( $values['value'] ); ?></td>
-								<?php if( $values['editable'] ): ?>
+						<?php if ( $values['value'] && isset( $values['type'] ) ) : ?>
+							<?php if ( isset( $values['editable'] ) && $this->current_user_can_modify_post( $post ) ) : ?>
+								<td class="value
```
alecgeatches: This seems to add some newlines in the class attribute which seems off. Could we pre-calculate `editable-value`, something like this?

```php
<?php $editable_class = $values['editable'] ? 'editable-value' : ''; ?>

// ...

<td class="value <?php echo esc_attr( $editable_class ) ?>"><?php echo esc_html( $values['value'] ); ?></td>
```

modules/custom-status/custom-status.php
```diff
@@ -459,747 +466,806 @@ function post_admin_header() {
...
-		$status_slug = sanitize_title( $status_name );
-		$status_description = stripslashes( wp_filter_nohtml_kses( trim( $_POST['status_description'] ) ) );
+			// Try to add the status
+			$status_args = array(
+				'description' => $status_description,
+				'slug' => $status_slug,
+			);
+			$return = $this->add_custom_status( $status_name, $status_args );
+			if ( is_wp_error( $return ) ) {
+				wp_die( esc_textarea( __( 'Could not add status: ', 'edit-flow' ) . $return->get_error_message() ) );
```
alecgeatches: Couple suggestions:

1. We can use `esc_html()` here which should be more restrictive. `add_custom_status()` seems to primarily return potential errors from `wp_insert_term()`, which returns basic string error messages: https://github.com/WordPress/wordpress-develop/blob/6.5/src/wp-includes/taxonomy.php#L2399-L2766
2. Just for improvement's sake we could do a `sprintf()` instead of string concatenation to allow a translator to reorder where the error appears in this message.


```suggestion
			/* translators: %s: error message */
			wp_die( esc_html( sprintf( __( 'Could not add status: %s', 'edit-flow' ), $return->get_error_message() ) ) );
```

modules/custom-status/custom-status.php
```diff
@@ -459,747 +466,806 @@ function post_admin_header() {
...
-			wp_die( $this->module->messages['status-missing'] );
+			// Try to add the new post status
+			$args = array(
+				'name' => $name,
+				'slug' => sanitize_title( $name ),
+				'description' => $description,
+			);
+			$return = $this->update_custom_status( $existing_status->term_id, $args );
+			if ( is_wp_error( $return ) ) {
+				wp_die( esc_textarea( __( 'Error updating post status.', 'edit-flow' ) ) );
```
alecgeatches: `esc_html()` should be fine and more restrictive here.

alecgeatches: Or, even better

```suggestion
				wp_die( esc_html__( 'Error updating post status.', 'edit-flow' ) );
```

modules/custom-status/custom-status.php
```diff
@@ -459,747 +466,806 @@ function post_admin_header() {
...
+			$status_description = isset( $_POST['description'] ) ? stripslashes( wp_filter_nohtml_kses( trim( $_POST['description'] ) ) ) : '';
 
-		// Check if name field was filled in
-		if ( empty( $status_name ) ) {
-			$change_error = new WP_Error( 'invalid', __( 'Please enter a name for the status.', 'edit-flow' ) );
-			die( $change_error->get_error_message() );
-		}
+			// Check if name field was filled in
+			if ( empty( $status_name ) ) {
+				$change_error = new WP_Error( 'invalid', esc_html__( 'Please enter a name for the status.', 'edit-flow' ) );
```
alecgeatches, line 1105: Nit: Escaping HTML in an error isn't usually necessary, especially as we escape it in the output line below. I assume this came from a linting error though, so maybe it made you do that.

alecgeatches: (same for the 5 `WP_Error`/`die()`s below)

modules/custom-status/custom-status.php
```diff
@@ -459,747 +466,806 @@ function post_admin_header() {
...
+				'name' => $status_name,
+				'description' => $status_description,
+				'slug' => $status_slug,
+			);
+			$return = $this->update_custom_status( $term_id, $args );
+			if ( ! is_wp_error( $return ) ) {
+				set_current_screen( 'edit-custom-status' );
+				$wp_list_table = new EF_Custom_Status_List_Table();
+				$wp_list_table->prepare_items();
+				echo wp_kses_post( $wp_list_table->single_row( $return ) );
```
alecgeatches, line 1151: Just noting - `wp_kses_post()` on a table row seems odd. Ideally `$wp_list_table->single_row()` should return sanitized HTML (it doesn't) and we should ignore the warning here. However, I tested this (happens during quick edits on the EditFlow Admin -> Custom Statuses page, then hover over a status and click "Quick Edit" and then "Update Status"), and it seems to work totally fine.

modules/custom-status/custom-status.php
```diff
@@ -459,747 +466,806 @@ function post_admin_header() {
...
+				die();
+			} else {
+				/* translators: 1: the status's name */
+				$change_error = new WP_Error( 'invalid', wp_kses( sprintf( __( 'Could not update the status: <strong>%s</strong>', 'edit-flow' ), $status_name ), 'strong' ) );
+				die( esc_html( $change_error->get_error_message() ) );
```
alecgeatches: We're double escaping the content here, with `wp_kses` and then `esc_html`. Moving the `wp_kses()` call into `die()` should fix that and stop the `<strong>` from being output as escaped:

```suggestion
				die( wp_kses( $change_error->get_error_message(), 'strong' ) );
```

modules/custom-status/custom-status.php
```diff
@@ -459,747 +466,806 @@ function post_admin_header() {
...
-		?>
+				<?php
+				wp_original_referer_field();
+				wp_nonce_field( 'edit-status' );
+				?>
 		<table class="form-table">
 			<tr class="form-field form-required">
 				<th scope="row" valign="top"><label for="name"><?php _e( 'Custom Status', 'edit-flow' ); ?></label></th>
-				<td><input name="name" id="name" type="text" value="<?php echo esc_attr( $name ); ?>" size="40" aria-required="true" <?php if( 'draft' === $status->slug ) echo 'readonly="readonly"' ?> />
+				<td><input name="name" id="name" type="text" value="<?php echo esc_attr( $name ); ?>" size="40" aria-required="true"
```
alecgeatches: Same suggestion as the other multi-line class, I'd recommend [pre-computing the extra attribute](https://github.com/Automattic/Edit-Flow/pull/734#discussion_r1655490459) and echoing, or using a ternary instead.

modules/custom-status/custom-status.php
```diff
@@ -1243,515 +1309,544 @@ function print_configure_view() {
 				<div class="col-wrap">
 				<div class="form-wrap">
 				<h3 class="nav-tab-wrapper">
-					<a href="<?php echo esc_url( $this->get_link() ); ?>" class="nav-tab<?php if ( !isset( $_GET['action'] ) || $_GET['action'] != 'change-options' ) echo ' nav-tab-active'; ?>"><?php _e( 'Add New', 'edit-flow' ); ?></a>
-					<a href="<?php echo esc_url( $this->get_link( array( 'action' => 'change-options' ) ) ); ?>" class="nav-tab<?php if ( isset( $_GET['action'] ) && $_GET['action'] == 'change-options' ) echo ' nav-tab-active'; ?>"><?php _e( 'Options', 'edit-flow' ); ?></a>
+					<a href="<?php echo esc_url( $this->get_link() ); ?>" class="nav-tab
```
alecgeatches: Same [inline class suggestion](https://github.com/Automattic/Edit-Flow/pull/734/files?diff=split&w=1#r1655521635) for this one.

modules/custom-status/custom-status.php
```diff
@@ -1243,515 +1309,544 @@ function print_configure_view() {
...
 				</form>
-				<?php else: ?>
-				<?php /** Custom form for adding a new Custom Status term **/ ?>
+				<?php else : ?>
+					<?php /** Custom form for adding a new Custom Status term **/ ?>
 					<form class="add:the-list:" action="<?php echo esc_url( $this->get_link() ); ?>" method="post" id="addstatus" name="addstatus">
 					<div class="form-field form-required">
 						<label for="status_name"><?php _e( 'Name', 'edit-flow' ); ?></label>
-						<input type="text" aria-required="true" size="20" maxlength="20" id="status_name" name="status_name" value="<?php if ( !empty( $_POST['status_name'] ) ) echo esc_attr( $_POST['status_name'] ) ?>" />
+						<input type="text" aria-required="true" size="20" maxlength="20" id="status_name" name="status_name" value="<?php /* phpcs:ignore Generic.ControlStructures.InlineControlStructure.NotAllowed */ if ( ! empty( $_POST['status_name'] ) ) echo esc_attr( $_POST['status_name'] ); ?>" />
```
alecgeatches: A ternary here might be easier:

```suggestion
						<input type="text" aria-required="true" size="20" maxlength="20" id="status_name" name="status_name" value="<?php echo ( empty( $_POST['status_name'] ) ? '' : esc_attr( $_POST['status_name'] ) ); ?>" />
```

alecgeatches: Same for the `phpcs:ignore` below

modules/dashboard/dashboard.php
```diff
@@ -11,354 +11,362 @@
...
-					$url = esc_url(get_edit_post_link( $post->ID ));
-					$title = esc_html($post->post_title);
+					$url = esc_url( get_edit_post_link( $post->ID ) );
+					$title = esc_html( $post->post_title );
 					?>
 					<li>
-						<h4><a href="<?php echo $url ?>" title="<?php _e('Edit this post', 'edit-flow') ?>"><?php echo $title; ?></a></h4>
-						<span class="ef-myposts-timestamp"><?php _e('This post was last updated on', 'edit-flow') ?> <?php echo get_the_time('F j, Y \\a\\t g:i a', $post) ?></span>
-					</li>	
+						<h4><a href="<?php echo esc_url( $url ); ?>" title="<?php _e( 'Edit this post', 'edit-flow' ); ?>"><?php echo esc_html( $title ); ?></a></h4>
```
alecgeatches, line 242: The `_e` on this line and the next should be `esc_html_e`s

alecgeatches:
```suggestion
						<h4><a href="<?php echo esc_url( $url ); ?>" title="<?php esc_attr_e( 'Edit this post', 'edit-flow' ); ?>"><?php echo esc_html( $title ); ?></a></h4>
```

modules/dashboard/dashboard.php
```diff
@@ -11,354 +11,362 @@
...
+						<span class="ef-myposts-timestamp"><?php _e( 'This post was last updated on', 'edit-flow' ); ?> <?php echo esc_html( get_the_time( 'F j, Y \\a\\t g:i a', $post ) ); ?></span>
```
alecgeatches:
```suggestion
						<span class="ef-myposts-timestamp"><?php esc_html_e( 'This post was last updated on', 'edit-flow' ); ?> <?php echo esc_html( get_the_time( 'F j, Y \\a\\t g:i a', $post ) ); ?></span>
```

modules/dashboard/dashboard.php
```diff
@@ -11,354 +11,362 @@
...
+					</li>
 				<?php endforeach; ?>
 			<?php else : ?>
-				<p><?php _e('Sorry! You\'re not subscribed to any posts!', 'edit-flow') ?></p>
+				<p><?php _e( 'Sorry! You\'re not subscribed to any posts!', 'edit-flow' ); ?></p>
```
alecgeatches:
```suggestion
				<p><?php esc_html_e( 'Sorry! You\'re not subscribed to any posts!', 'edit-flow' ); ?></p>
```

modules/editorial-comments/editorial-comments.php
```diff
@@ -6,117 +6,118 @@
...
-		wp_enqueue_style( 'edit-flow-editorial-comments-css', $this->module_url . 'lib/editorial-comments.css', false, EDIT_FLOW_VERSION, 'all' );
+			wp_enqueue_style( 'edit-flow-editorial-comments-css', $this->module_url . 'lib/editorial-comments.css', false, EDIT_FLOW_VERSION, 'all' );
 
-		$thread_comments = (int) get_option('thread_comments');
-		?>
+			$thread_comments = (int) get_option( 'thread_comments' );
+			?>
 		<script type="text/javascript">
-			var ef_thread_comments = <?php echo ($thread_comments) ? $thread_comments : 0; ?>;
+			var ef_thread_comments = <?php echo esc_html( ( $thread_comments ) ? $thread_comments : 0 ); ?>;
```
alecgeatches: Minor, could use `intval` for sanitization here to be more strict and specific:

```suggestion
			var ef_thread_comments = <?php echo $thread_comments ? intval( $thread_comments ) : 0; ?>;
```

modules/editorial-comments/editorial-comments.php
```diff
@@ -189,261 +190,283 @@ function the_comment_form( ) {
 
 			<p id="ef-replysubmit">
 				<a class="ef-replysave button-primary alignright" href="#comments-form">
-					<span id="ef-replybtn"><?php _e('Submit Response', 'edit-flow') ?></span>
+					<span id="ef-replybtn"><?php _e( 'Submit Response', 'edit-flow' ); ?></span>
```
alecgeatches:
```suggestion
					<span id="ef-replybtn"><?php esc_html_e( 'Submit Response', 'edit-flow' ); ?></span>
```

modules/editorial-comments/editorial-comments.php
```diff
@@ -189,261 +190,283 @@ function the_comment_form( ) {
...
 
-			$sep = ' ';
-			$i = 0;
-			foreach ( $actions as $action => $link ) {
-				++$i;
-				// Reply and quickedit need a hide-if-no-js span
-				if ( 'reply' == $action || 'quickedit' == $action )
-					$action .= ' hide-if-no-js';
+			// It's already been escaped above.
+			// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
alecgeatches: FYI `phpcs` has an inline way to add comments for a warning:

```suggestion
			// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped -- It's already been escaped above
```

In the [docs under **Ignoring Parts of a File**](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#ignoring-parts-of-a-file):

> If required, you can add a note explaining why sniffs are being disable and re-enabled by using the -- separator.
>
> ```php
> // phpcs:disable PEAR,Squiz.Arrays -- this isn't our code
> $foo = [1,2,3];
> bar($foo,true);
> // phpcs:enable PEAR.Functions.FunctionCallSignature -- check function calls again
> bar($foo,false);
> // phpcs:enable -- this is out code again, so turn everything back on
> ```

I typically like putting an explanation on all `phpcs:ignore`/`phpcs:disable`s, but that's probably out of scope for all the changes you're making here.

modules/editorial-metadata/editorial-metadata.php
```diff
@@ -14,311 +14,312 @@
...
+									'min-width: 8em;',
+								);
+								break;
+						}
+					}
+					// Allow users to filter out rules if there's something wonky
+					$css_rules = apply_filters( 'ef_editorial_metadata_manage_posts_css_rules', $css_rules );
+					echo "<style type=\"text/css\">\n";
+					foreach ( (array) $css_rules as $css_property => $rules ) {
+						echo wp_kses_post( $css_property . ' {' . implode( ' ', $rules ) . "}\n " );
```
alecgeatches: Might just be worth echoing without sanitization, and `phpcs:ignore`ing this line. `wp_kses_post()` probably can break valid CSS, and I don't think makes sense for filter-provided CSS.

modules/editorial-metadata/editorial-metadata.php
```diff
@@ -327,1114 +328,1203 @@ function register_taxonomy() {
...
+
+		/**
+		 * Displays HTML output for Editorial Metadata post meta box
+		 *
+		 * @param object $post Current post
+		 */
+		public function display_meta_box( $post ) {
+			echo "<div id='" . esc_attr( self::metadata_taxonomy ) . "_meta_box'>";
+			// Add nonce for verification upon save
+			echo "<input type='hidden' name='" . esc_attr( self::metadata_taxonomy ) . "_nonce' value='" . esc_js( wp_create_nonce( 'ef-save-metabox' ) ) . "' />";
```
alecgeatches: I think `esc_attr()` is more appropriate for a nonce value, which is just a alpha-numeric string.

```suggestion
			echo "<input type='hidden' name='" . esc_attr( self::metadata_taxonomy ) . "_nonce' value='" . esc_attr( wp_create_nonce( 'ef-save-metabox' ) ) . "' />";
```

modules/editorial-metadata/editorial-metadata.php
```diff
@@ -327,1114 +328,1203 @@ function register_taxonomy() {
...
-	 *
-	 * @param array $calendar_fields Additional data fields to include on the calendar
-	 * @param int $post_id Unique ID for the post data we're building
-	 * @return array $calendar_fields Calendar fields with our viewable Editorial Metadata added
-	 */
-	function filter_calendar_item_fields( $calendar_fields, $post_id ) {
+			// or that the term name doesn't map to an existing term's slug
+			$search_term = $this->get_editorial_metadata_term_by( 'slug', sanitize_title( $metadata_name ) );
+			if ( is_object( $search_term ) && $search_term->term_id != $existing_term->term_id ) {
+				$change_error = new WP_Error( 'invalid', esc_html__( 'Name conflicts with slug for another term. Please choose again.', 'edit-flow' ) );
```
alecgeatches, line 1331: Double-escape here, can use use `__()` and let `die()` escape below

modules/editorial-metadata/editorial-metadata.php
```diff
@@ -327,1114 +328,1203 @@ function register_taxonomy() {
...
+			if ( ! is_wp_error( $return ) ) {
+				set_current_screen( 'edit-editorial-metadata' );
+				$wp_list_table = new EF_Editorial_Metadata_List_Table();
+				$wp_list_table->prepare_items();
+				echo wp_kses_post( $wp_list_table->single_row( $return ) );
+				die();
+			} else {
+				/* Translators: 1: the name of the term that could not be found */
+				$change_error = new WP_Error( 'invalid', wp_kses( sprintf( __( 'Could not update the term: <strong>%s</strong>', 'edit-flow' ), $metadata_name ), 'strong' ) );
+				die( esc_html( $change_error->get_error_message() ) );
```
alecgeatches: This'll output a literal `<strong>` insteag of the HTML tag due to double-escaping. Can move the `wp_kses()` call into `die()` like this other comment: https://github.com/Automattic/Edit-Flow/pull/734/files/0c6d169d72a8d91145d034d987371c97447b6869#r1655519700

modules/editorial-metadata/editorial-metadata.php
```diff
@@ -1470,57 +1560,81 @@ function print_configure_view() {
...
+		<?php else : ?>
+			<?php /** If not in full-screen edit term mode, we can create new terms or change options **/ ?>
 		<div id="col-left">
-			<div class="col-wrap">	
+			<div class="col-wrap">
 			<div class="form-wrap">
 			<h3 class="nav-tab-wrapper">
-				<a href="<?php echo esc_url( add_query_arg( array( 'page' => $this->module->settings_slug ), get_admin_url( null, 'admin.php' ) ) ); ?>" class="nav-tab<?php if ( !isset( $_GET['action'] ) || $_GET['action'] != 'change-options' ) echo ' nav-tab-active'; ?>"><?php _e( 'Add New', 'edit-flow' ); ?></a>
-				<a href="<?php echo esc_url( add_query_arg( array( 'page' => $this->module->settings_slug, 'action' => 'change-options' ), get_admin_url( null, 'admin.php' ) ) ); ?>" class="nav-tab<?php if ( isset( $_GET['action'] ) && $_GET['action'] == 'change-options' ) echo ' nav-tab-active'; ?>"><?php _e( 'Options', 'edit-flow' ); ?></a>
+				<a href="<?php echo esc_url( add_query_arg( array( 'page' => $this->module->settings_slug ), get_admin_url( null, 'admin.php' ) ) ); ?>" class="nav-tab
```
alecgeatches: Another opportunity to [pre-compute class names](https://github.com/Automattic/Edit-Flow/pull/734#discussion_r1655490459) to avoid this layout.

modules/editorial-metadata/editorial-metadata.php
```diff
@@ -1470,57 +1560,81 @@ function print_configure_view() {
...
+					'page' => $this->module->settings_slug,
+					'action' => 'change-options',
+				), get_admin_url( null, 'admin.php' ) ) );
+				?>
+							" class="nav-tab
+							<?php
+							if ( isset( $_GET['action'] ) && 'change-options' == $_GET['action'] ) {
+														echo ' nav-tab-active';}
+							?>
+"><?php _e( 'Options', 'edit-flow' ); ?></a>
```
alecgeatches:
```suggestion
"><?php esc_hml_e( 'Options', 'edit-flow' ); ?></a>
```

modules/story-budget/story-budget.php
```diff
@@ -328,22 +334,22 @@ function story_budget() {
...
 					echo '</div>';
 				?>
 				<style>
 					<?php
-					  for ( $i = 1; $i <= $this->max_num_columns; ++$i ) {
+					for ( $i = 1; $i <= $this->max_num_columns; ++$i ) {
 						?>
 					.columns-number-<?php echo (int) $i; ?> .postbox {
-						flex-basis: <?php echo  99 / $i ?>%;
+						flex-basis: <?php echo 99 / /* phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped */ $i; ?>%;
```
alecgeatches:
```suggestion
						flex-basis: <?php echo floatval( 99 / $i ); ?>%;
```

should work for sanitization too.

modules/user-groups/user-groups.php
```diff
@@ -1,573 +1,613 @@
...
+			}
+
+			// Try to add the Usergroup
+			$args = array(
+				'name' => $name,
+				'description' => $description,
+			);
+			$usergroup = $this->add_usergroup( $args );
+			if ( is_wp_error( $usergroup ) ) {
+				wp_die( esc_textarea( __( 'Error adding usergroup.', 'edit-flow' ) ) );
```
alecgeatches:
```suggestion
				wp_die( esc_html__('Error adding usergroup.', 'edit-flow' ) );
```

modules/user-groups/user-groups.php
```diff
@@ -1,573 +1,613 @@
...
+			$args = array(
+				'name' => $name,
+				'description' => $description,
+			);
+			// Gracefully handle the case where all users have been unsubscribed from the user group
+			$users = isset( $_POST['usergroup_users'] ) ? (array) $_POST['usergroup_users'] : array();
+			$users = array_map( 'intval', $users );
+			$usergroup = $this->update_usergroup( $existing_usergroup->term_id, $args, $users );
+			if ( is_wp_error( $usergroup ) ) {
+				wp_die( esc_textarea( __( 'Error updating user group.', 'edit-flow' ) ) );
```
alecgeatches:
```suggestion
				wp_die( esc_html__( 'Error updating user group.', 'edit-flow' ) );
```

modules/user-groups/user-groups.php
```diff
@@ -1,573 +1,613 @@
...
+
+			$name = isset( $_POST['name'] ) ? sanitize_text_field( trim( $_POST['name'] ) ) : '';
+			$description = isset( $_POST['description'] ) ? stripslashes( wp_filter_nohtml_kses( trim( $_POST['description'] ) ) ) : '';
+
+			/**
+			 * Form validation for editing Usergroup
+			 */
+			// Check if name field was filled in
+			if ( empty( $name ) ) {
+				$change_error = new WP_Error( 'invalid', esc_html__( 'Please enter a name for the user group.', 'edit-flow' ) );
```
alecgeatches, line 482:
```suggestion
				$change_error = new WP_Error( 'invalid', __( 'Please enter a name for the user group.', 'edit-flow' ) );
```

modules/user-groups/user-groups.php
```diff
@@ -1,573 +1,613 @@
...
+				die( esc_html( $change_error->get_error_message() ) );
+			}
+			// Check that the name doesn't exceed 40 chars
+			if ( strlen( $name ) > 40 ) {
+				$change_error = new WP_Error( 'invalid', esc_html__( 'User group name cannot exceed 40 characters. Please try a shorter name.' ) );
```
alecgeatches, line 487:
```suggestion
				$change_error = new WP_Error( 'invalid', __( 'User group name cannot exceed 40 characters. Please try a shorter name.' ) );
```

alecgeatches: (same for the two below)

modules/user-groups/user-groups.php
```diff
@@ -1,573 +1,613 @@
...
+			if ( ! is_wp_error( $return ) ) {
+				set_current_screen( 'edit-usergroup' );
+				$wp_list_table = new EF_Usergroups_List_Table();
+				$wp_list_table->prepare_items();
+				echo wp_kses_post( $wp_list_table->single_row( $return ) );
+				die();
+			} else {
+				// translators: %s is the name of the user group
+				$change_error = new WP_Error( 'invalid', sprintf( __( 'Could not update the user group: <strong>%s</strong>', 'edit-flow' ), $name ) );
+				die( esc_html( $change_error->get_error_message() ) );
```
alecgeatches: An echoed `<strong>` tag here. Can do [the `wp_kses()` thing here](https://github.com/Automattic/Edit-Flow/pull/734#discussion_r1655519700).

modules/user-groups/user-groups.php
```diff
@@ -579,83 +619,98 @@ function print_configure_view() {
...
+				var ef_confirm_delete_usergroup_string = "<?php _e( 'Are you sure you want to delete the user group?', 'edit-flow' ); ?>";
 			</script>
 			<div id="col-right"><div class="col-wrap">
 				<?php $wp_list_table->display(); ?>
 			</div></div>
 			<div id="col-left"><div class="col-wrap"><div class="form-wrap">
 				<h3 class="nav-tab-wrapper">
-					<a href="<?php echo esc_url( $this->get_link() ); ?>" class="nav-tab<?php if ( !isset( $_GET['action'] ) || $_GET['action'] != 'change-options' ) echo ' nav-tab-active'; ?>"><?php _e( 'Add New', 'edit-flow' ); ?></a>
-					<a href="<?php echo esc_url( $this->get_link( array( 'action' => 'change-options' ) ) ); ?>" class="nav-tab<?php if ( isset( $_GET['action'] ) && $_GET['action'] == 'change-options' ) echo ' nav-tab-active'; ?>"><?php _e( 'Options', 'edit-flow' ); ?></a>
+					<a href="<?php echo esc_url( $this->get_link() ); ?>" class="nav-tab
```
alecgeatches: A split class here that could be precomputed or ternaryized.

modules/user-groups/user-groups.php
```diff
@@ -579,83 +619,98 @@ function print_configure_view() {
...
 				</form>
-				<?php else: ?>
-				<?php /** Custom form for adding a new Usergroup **/ ?>
+				<?php else : ?>
+					<?php /** Custom form for adding a new Usergroup **/ ?>
 					<form class="add:the-list:" action="<?php echo esc_url( $this->get_link() ); ?>" method="post" id="addusergroup" name="addusergroup">
 					<div class="form-field form-required">
 						<label for="name"><?php _e( 'Name', 'edit-flow' ); ?></label>
-						<input type="text" aria-required="true" id="name" name="name" maxlength="40" value="<?php if ( !empty( $_POST['name'] ) ) echo esc_attr( $_POST['name'] ); ?>" />
+						<input type="text" aria-required="true" id="name" name="name" maxlength="40" value="<?php /* phpcs:ignore Generic.ControlStructures.InlineControlStructure.NotAllowed */ if ( ! empty( $_POST['name'] ) ) echo esc_attr( $_POST['name'] ); ?>"
```
alecgeatches:
```suggestion
						<input type="text" aria-required="true" id="name" name="name" maxlength="40" value="<?php echo esc_attr( isset( $_POST['name'] ) && ! empty( $_POST['name'] ) ? $_POST['name'] : '' );
```

alecgeatches: Same suggestion to ternary the other `phpcs:ignore` below.

modules/user-groups/user-groups.php
```diff
@@ -667,571 +722,600 @@ function user_profile_page() {
...
+				'list_id' => 'ef-following_usergroups',
+				'input_id' => 'following_usergroups',
+			);
+
+			$parsed_args = wp_parse_args( $args, $defaults );
+			extract( $parsed_args, EXTR_SKIP );
+			$usergroups = $this->get_usergroups();
+			if ( empty( $usergroups ) ) {
 				?>
+			<p><?php _e( 'No user groups were found.', 'edit-flow' ); ?> <a href="<?php echo esc_url( $this->get_link() ); ?>" title="<?php _e( 'Add a new user group. Opens new window.', 'edit-flow' ); ?>" target="_blank"><?php _e( 'Add a User Group', 'edit-flow' ); ?></a></p>
```
alecgeatches:
```suggestion
			<p><?php esc_html_e( 'No user groups were found.', 'edit-flow' ); ?> <a href="<?php echo esc_url( $this->get_link() ); ?>" title="<?php esc_attr_e( 'Add a new user group. Opens new window.', 'edit-flow' ); ?>" target="_blank"><?php esc_html_e( 'Add a User Group', 'edit-flow' ); ?></a></p>
```

modules/user-groups/user-groups.php
```diff
@@ -667,571 +722,600 @@ function user_profile_page() {
...
+				<?php
+				foreach ( $usergroups as $usergroup ) {
+					$checked = ( in_array( $usergroup->term_id, $selected ) ) ? ' checked="checked"' : '';
+					?>
 				<li>
-					<label for="<?php echo $input_id . esc_attr( $usergroup->term_id ); ?>" title="<?php echo esc_attr($usergroup->description) ?>">
+					<label for="<?php echo esc_attr( $input_id ) . esc_attr( $usergroup->term_id ); ?>" title="<?php echo esc_attr( $usergroup->description ); ?>">
 						<div class="ef-user-subscribe-actions">
-							<input type="checkbox" id="<?php echo esc_attr( $input_id . $usergroup->term_id ) ?>" name="<?php echo esc_attr( $input_id )?>[]" value="<?php echo esc_attr( $usergroup->term_id ) ?>"<?php echo $checked ?> />
+							<input type="checkbox" id="<?php echo esc_attr( $input_id . $usergroup->term_id ); ?>" name="<?php echo esc_attr( $input_id ); ?>[]" value="<?php echo esc_attr( $usergroup->term_id ); ?>"<?php echo wp_kses_post( $checked ); ?> />
```
alecgeatches: WordPress has a [built-in `checked()` method](https://developer.wordpress.org/reference/functions/checked/) that should let us avoid using `wp_kses_post()`, and also avoid precomputing the HTML value of `$checked`:

```suggestion
							<input type="checkbox" id="<?php echo esc_attr( $input_id . $usergroup->term_id ); ?>" name="<?php echo esc_attr( $input_id ); ?>[]" value="<?php echo esc_attr( $usergroup->term_id ); ?>"<?php checked( in_array( $usergroup->term_id, $selected ) ); ?> />
```

alecgeatches (changes requested):
@ingeniumed Thank you so much for doing these changes! There were a ton of really valuable updates to escaping functions, translator strings, and `isset` checks, as well as a lot of code-quality updates like adding braced and correcting code spacing. I found [turning whitespace off](https://github.com/Automattic/Edit-Flow/pull/734/files?w=1) (settings gear -> hide whitespace) in the diff was very necessary for review. Note I also tested features manually in this branch to ensure nothing obvious was broken, and everything loaded and ran well.

I left comments on specific issues I noticed above.

There was one big change we may want to reconsider ([mentioned once here](https://github.com/Automattic/Edit-Flow/pull/734#discussion_r1655478882)), and that's the switch from `date()` to `gmdate()`. This was done in a lot of places. `gmdate()` is the correct function to use generally speaking, but these functions return different values.

Unless we're really careful to migrate date calculations correctly, this could cause scheduled posts and other date-related features (like the calendar) to suddenly change and be off by a few hours. I think users typically expect WordPress to reflect the date and timezone set in Admin -> Settings -> General, and modifying our core date function without careful testing could cause unexpected behavior in multiple features.

I don't know of any specific place of where this will break, but I think I could find some if you'd like a more concrete example.

What do you think?

ingeniumed:
> @ingeniumed Thank you so much for doing these changes! There were a ton of really valuable updates to escaping functions, translator strings, and `isset` checks, as well as a lot of code-quality updates like adding braced and correcting code spacing. I found [turning whitespace off](https://github.com/Automattic/Edit-Flow/pull/734/files?w=1) (settings gear -> hide whitespace) in the diff was very necessary for review. Note I also tested features manually in this branch to ensure nothing obvious was broken, and everything loaded and ran well.
> 
> I left comments on specific issues I noticed above.
> 
> There was one big change we may want to reconsider ([mentioned once here](https://github.com/Automattic/Edit-Flow/pull/734#discussion_r1655478882)), and that's the switch from `date()` to `gmdate()`. This was done in a lot of places. `gmdate()` is the correct function to use generally speaking, but these functions return different values.
> 
> Unless we're really careful to migrate date calculations correctly, this could cause scheduled posts and other date-related features (like the calendar) to suddenly change and be off by a few hours. I think users typically expect WordPress to reflect the date and timezone set in Admin -> Settings -> General, and modifying our core date function without careful testing could cause unexpected behavior in multiple features.
> 
> I don't know of any specific place of where this will break, but I think I could find some if you'd like a more concrete example.
> 
> What do you think?

Thanks for reviewing it, this isn't an easy PR to review so I appreciate it.

I think out of safety, I'll revert that change and exclude that rule from phpcs ignore for now. In a follow up PR, we can do rigorous testing and get that in. While I did test it briefly, and it worked well I don't think we should risk letting it go in in this giant PR. You are 100% right on it imo

modules/notifications/notifications.php
```diff
@@ -836,7 +836,7 @@ public function get_notification_footer( $post ) {
 			$body .= sprintf( __( 'You are receiving this email because you are subscribed to "%s".', 'edit-flow' ), ef_draft_or_post_title( $post->ID ) );
 			$body .= "\r\n";
 			/* translators: 1: date */
-			$body .= sprintf( __( 'This email was sent %s.', 'edit-flow' ), gmdate( 'r' ) );
+			$body .= sprintf( __( 'This email was sent %s.', 'edit-flow' ), date( 'r' ) );
```
ingeniumed, line 839: Realized this was switched in a previous PR, so I have reverted it.

modules/calendar/calendar.php
ingeniumed: Reverted this as discussed

modules/custom-status/custom-status.php
ingeniumed: It's def the linter forcing me to do it, and I have the same qualm with it - why do it if it's being escaped already at the bottom? I wanted to reduce the amount of ignores that I had in place too tbh, so I just went ahead and did it

modules/custom-status/custom-status.php
ingeniumed: Yip, agreed on that. That's how I had tried it, but the resulting ignore wherever this table class is used wasn't too well liked by me. I'd like to minimize the ignores so I ended up doing this. 

That said, I'm thinking that we can see during our feature development if we are going to keep this class or do it differently perhaps.

modules/custom-status/custom-status.php
ingeniumed: Good catch, the linter complained about both of em and I ended up messing up the right escape to use.

modules/editorial-metadata/editorial-metadata.php
ingeniumed: This one's the same as custom status - it's due to the linting rules raising a concern here. Would rather not have ignores all over so I just ended up double escaping

modules/user-groups/user-groups.php
ingeniumed: This one's like this for the same reason as the other modules. The linter forces it and I'd rather not have ignores all over the place

modules/user-groups/user-groups.php
ingeniumed: This one's like this for the same reason as the other modules. The linter forces it and I'd rather not have ignores all over the place

modules/user-groups/user-groups.php
ingeniumed: It's interesting because phpcs still raised an error for this. Lemme try it again

modules/calendar/calendar.php
ingeniumed: Done

modules/custom-status/custom-status.php
ingeniumed: Done

modules/custom-status/custom-status.php
ingeniumed: Done

modules/custom-status/custom-status.php
ingeniumed: Done

modules/custom-status/custom-status.php
ingeniumed: Done

modules/custom-status/custom-status.php
ingeniumed: Done

modules/custom-status/custom-status.php
ingeniumed: Done

modules/dashboard/dashboard.php
ingeniumed: Done

modules/dashboard/dashboard.php
ingeniumed: Done

modules/editorial-comments/editorial-comments.php
ingeniumed: Done

modules/editorial-comments/editorial-comments.php
ingeniumed: Done

modules/editorial-comments/editorial-comments.php
ingeniumed: Done

modules/editorial-metadata/editorial-metadata.php
ingeniumed: Done

modules/editorial-metadata/editorial-metadata.php
ingeniumed: Done

modules/editorial-metadata/editorial-metadata.php
ingeniumed: Done

modules/editorial-metadata/editorial-metadata.php
ingeniumed: Done

modules/editorial-metadata/editorial-metadata.php
ingeniumed: Done

modules/story-budget/story-budget.php
ingeniumed: Done

modules/user-groups/user-groups.php
ingeniumed: Done

modules/user-groups/user-groups.php
ingeniumed: Done

modules/user-groups/user-groups.php
ingeniumed: Done

modules/user-groups/user-groups.php
ingeniumed: Done

modules/user-groups/user-groups.php
ingeniumed: Done

modules/user-groups/user-groups.php
ingeniumed: Done

modules/user-groups/user-groups.php
ingeniumed: Done

alecgeatches (approved):
Looking good! Thank you for addressing the comments above.

PR #736, opened 2024-06-27
Refactor custom status module views

alecgeatches:
## Description

This PR is currently based on @ingeniumed's work in https://github.com/Automattic/Edit-Flow/pull/734. Please review after that PR has been merged.

---

Some cleanup for the custom status module, as initial work to extend it. There were a few main changes:

1. The `print_configure_view()` function had a lot of responsibilities, embedded HTML, and indentation. I moved the two HTML sections (configure page, edit status page) into separate view files and simplified the function.
2. Fixed some PHPCS warnings from using `$_GET` and `$_POST` variables in `print_configure_view()` with some rewrites and some `phpcs:ignore`s.
3. Updated i18n functions to their escaped versions (`_e()` → `esc_html_e()`) where appropriate.
4. Moved one string (`ef_confirm_delete_status_string`) from an embeded `<script>` tag to `wp_localize_script()`.
5. Updated all `array()`s in the file to short syntax (`[]`). I did this automatically with [PHP-CS-Fixer](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer):

    ```bash
    php-cs-fixer fix modules/custom-status/custom-status.php --rules='{"array_syntax": {"syntax": "short"}}'
    ```

6. A lot of whitespace alignment changes were automatically made by my editor.

## Steps to Test

1. Check out PR.
1. Go to Admin -> Edit Flow -> Custom Statuses.
1. Ensure adding and editing custom statuses works as expected.

### DIFF
--- a/modules/custom-status/custom-status.php
+++ b/modules/custom-status/custom-status.php
@@ -15,7 +15,7 @@ if ( ! class_exists( 'EF_Custom_Status' ) ) {
 
 		public $module;
 
-		private $custom_statuses_cache = array();
+		private $custom_statuses_cache = [];
 
 		// This is taxonomy name used to store all our custom statuses
 		// phpcs:ignore Generic.NamingConventions.UpperCaseConstantName.ClassConstantNotUpperCase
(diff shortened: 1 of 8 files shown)


modules/custom-status/custom-status.php
```diff
@@ -1229,119 +1233,42 @@ public function settings_validate( $new_options ) {
...
-			<?php endif; ?>
-			<?php
+				$is_nonce_valid = isset( $_POST['_wpnonce'] ) && wp_verify_nonce( wp_strip_all_tags( $_POST['_wpnonce'] ), 'edit-status' );
+
+				if ( $is_nonce_valid ) {
+					$name        = ( isset( $_POST['name'] ) ) ? wp_strip_all_tags( $_POST['name'] ) : $custom_status->name;
+					$description = ( isset( $_POST['description'] ) ) ? wp_strip_all_tags( $_POST['description'] ) : $custom_status->description;
+				}
+
+				include __DIR__ . '/views/edit-status.php';
```
hanifn: Unless it's a shared/common view file, I usually default to using `include_once` to ensure no one accidentally include the file elsewhere. Would that fit the scenario here?

modules/custom-status/custom-status.php
```diff
@@ -1229,119 +1233,42 @@ public function settings_validate( $new_options ) {
...
+			} else {
+				$custom_status_list_table = new EF_Custom_Status_List_Table();
+				$custom_status_list_table->prepare_items();
+				include __DIR__ . '/views/configure.php';
```
hanifn: Same as above

modules/custom-status/custom-status.php
ingeniumed: I like that, better to not keep including it again and again.

modules/custom-status/custom-status.php
ingeniumed: I like that, better to not keep including it again and again.

modules/custom-status/views/configure.php
```diff
@@ -0,0 +1,60 @@
...
+				<a href="<?php echo esc_url( $this->get_link() ); ?>" class="nav-tab <?php echo esc_attr( $add_new_nav_class ); ?>"><?php esc_html_e( 'Add New', 'edit-flow' ); ?></a>
+				<?php $options_nav_class = 'change-options' === $action ? 'nav-tab-active' : ''; ?>
+				<a href="<?php echo esc_url( $this->get_link( array( 'action' => 'change-options' ) ) ); ?>" class="nav-tab <?php echo esc_attr( $options_nav_class ); ?>"><?php esc_html_e( 'Options', 'edit-flow' ); ?></a>
+			</h3>
+
+			<?php if ( 'change-options' === $action ) { ?>
+			<form class="basic-settings" action="<?php echo esc_url( $this->get_link( array( 'action' => 'change-options' ) ) ); ?>" method="post">
+				<?php settings_fields( $this->module->options_group_name ); ?>
+				<?php do_settings_sections( $this->module->options_group_name ); ?>
+				<?php echo '<input id="edit_flow_module_name" name="edit_flow_module_name" type="hidden" value="' . esc_attr( $this->module->name ) . '" />'; ?>
```
hanifn: I think we can avoid `echo`ing out the entire input field if we instead just echo out the module name.
Like this:
```suggestion
				<input id="edit_flow_module_name" name="edit_flow_module_name" type="hidden" value="<?php echo esc_attr( $this->module->name ); ?>" />
```

modules/custom-status/views/configure.php
```diff
@@ -0,0 +1,60 @@
...
+				</div>
+
+				<div class="form-field">
+					<label for="status_description"><?php esc_html_e( 'Description', 'edit-flow' ); ?></label>
+					<textarea cols="40" rows="5" id="status_description" name="status_description"><?php echo ( empty( $_POST['status_description'] ) ? '' : esc_textarea( $_POST['status_description'] ) ); ?></textarea>
+					<?php $edit_flow->settings->helper_print_error_or_description( 'description', __( 'The description is primarily for administrative use, to give you some context on what the custom status is to be used for.', 'edit-flow' ) ); ?>
+				</div>
+
+				<?php wp_nonce_field( 'custom-status-add-nonce' ); ?>
+				<?php echo '<input id="action" name="action" type="hidden" value="add-new" />'; ?>
```
hanifn: Not really necessary to echo this line out I think
```suggestion
				<input id="action" name="action" type="hidden" value="add-new" />
```

hanifn (approved):
Code changes looks good to me and everything seems to be working as expected 👍 
🚢

PR #737, opened 2024-06-28
Update the analytics flag

ingeniumed:
## Description

This PR will remove the analytics dropdown and option in favour of enabling it for VIP Production sites by default only. For everyone else it will be **disabled**. 

There's a filter provided to disable or enable it. This can be used by VIP customers to disable it or by .org users to opt into analytics if they want it. 

Once we add in tracks integration, we can link the VIP privacy policy in the readme. I want to emphasize that this will be enabled for VIP customers only, for everyone else it's disabled by default.


hanifn (approved):
LGTM! 🚢

PR #738, opened 2024-07-01
Lint the common modules

ingeniumed:
## Description

This will lint the two common modules that were skipped out on, in the previous PR. It's mainly spacing changes really so it's safe to do so and I ensure this won't be ignored in the future.

## Steps to Test

- Use the screen options on the dashboard to enable/disable edit flow widgets
- Use the help screen to ensure everything renders correctly
- Preview posts with custom statuses

### DIFF
--- a/common/php/util.php
+++ b/common/php/util.php
@@ -1,6 +1,6 @@
 <?php
 
-if( ! function_exists( 'ef_draft_or_post_title' ) ) :
+if ( ! function_exists( 'ef_draft_or_post_title' ) ) :
 	/**
 	 * Copy of core's _draft_or_post_title without the filters
 	 *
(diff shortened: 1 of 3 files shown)


common/php/util.php
```diff
@@ -40,20 +40,22 @@ function _ef_wp_link_page( $i, $custom_statuses ) {
...
 		} else {
 			// Check for all custom post statuses, not just draft & pending
-			if ( '' == get_option('permalink_structure') || in_array($post->post_status, array_merge( $custom_statuses, array( 'pending' ) ) ) )
+			if ( '' == get_option( 'permalink_structure' ) || in_array( $post->post_status, array_merge( $custom_statuses, array( 'pending' ) ) ) ) {
 				$url = add_query_arg( 'page', $i, get_permalink() );
-			elseif ( 'page' == get_option('show_on_front') && get_option('page_on_front') == $post->ID )
-				$url = trailingslashit(get_permalink()) . user_trailingslashit("$wp_rewrite->pagination_base/" . $i, 'single_paged');
-			else
-				$url = trailingslashit(get_permalink()) . user_trailingslashit($i, 'single_paged');
+			} else if ( 'page' == get_option( 'show_on_front' ) && get_option( 'page_on_front' ) == $post->ID ) {
```
hanifn, line 48: Have we ever decided between `else if` vs `elseif`? Because I think there have been a PR or 2 before this where we've changed `else if` to `elseif`. 🤔

hanifn (approved):
Looks good. Just have a non-blocking question.

common/php/util.php
ingeniumed: Don't think so, though I'm not pedantic about that. I've always been used to writing `else if` hence why I stuck with it

PR #743, opened 2025-04-01
Fix PHP notice with null returned value from get_post_status_object()

srtfisher:
As titled.

### DIFF
--- a/modules/custom-status/custom-status.php
+++ b/modules/custom-status/custom-status.php
@@ -757,6 +757,10 @@ if ( ! class_exists( 'EF_Custom_Status' ) ) {
 
 			$post_status = get_post_status_object( get_post_status( $post->ID ) );
 
+			if ( ! is_object( $post_status ) ) {
+				return $post_states;
+			}
+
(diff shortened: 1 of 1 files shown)


GaryJones:
Thanks for the PR! You're right that `get_post_status_object()` can return `null` and we should handle that case.

**Why this happens:**

`get_post_status_object()` returns `null` when the status isn't registered. This can occur in several scenarios:
- Posts imported from another system with custom statuses that aren't registered in the current install
- Posts orphaned after a plugin (like Edit Flow) is deactivated, leaving posts with statuses like `pitch` or `assigned` that no longer exist
- Race conditions during plugin activation/deactivation
- Database inconsistencies

**Issue with the current fix:**

The current code returns early without a value:

```php
if ( ! is_object( $post_status ) ) {
    return;
}
```

This effectively returns `null`, which breaks the `display_post_states` filter chain. The filter expects `$post_states` (an array) to be returned so other filters can continue processing. Returning `null` would cause issues for any subsequent filters.

**Suggested fix:**

```php
if ( ! is_object( $post_status ) ) {
    return $post_states;
}
```

This preserves the filter chain while safely skipping our custom status display when the status object isn't available.

PR #747, opened 2025-06-23
Fix -- PHP Warning:  Attempt to read property "post_type" on string i…

awps:
## Description

Not sure when this is emitted, but sometimes I see this warning and it is flooding the logs. Probably the problem is somewhere else, or is not a problem at all, but this should mute this warning: 

```
PHP Warning:  Attempt to read property "post_type" on string in .../edit-flow/modules/custom-status/custom-status.php on line 1476
```


GaryJones:
Thank you for this fix, @dev4starter! 🙏

I've incorporated your fix into #834 with a couple of improvements:
- Rebased onto the `develop` branch (this PR was against `master` which doesn't run CI)
- Used `instanceof WP_Post` instead of `is_object()` for stricter type checking
- Fixed indentation to use tabs matching surrounding code

Your contribution is credited in the commit message and PR description.

PR #751, opened 2025-08-05
Update user-groups.php to avoid resaving usergroup for user already in group

jerclarke:
## Description

This small change represents an enormous performance improvement when there are a lot of users and groups. 

Without this check if the user is already in the group it gets re-saved anyway. If the user is in many groups this represents a lot of wasted time, and on our site it was slowing down user saving to the point where it felt broken. 

I see no risk in this check being added to avoid re-saving a user who is already in a group.  I've been running this as a hotfix in my plugin for years and saving users to groups works fine.

Thank you for considering it.

## Steps to Test

Outline the steps to test and verify the PR here.

* Add and remove users from groups
* Make sure everything works fine
* Enjoy the speedup caused by not resaving when you don't have to


GaryJones:

PR #752, opened 2025-08-25
Fix numeric HTML ID in calendar.php

jerclarke:
HTML IDs aren't allowed to start with a number, but these `td.day-unit` elements all had IDs like `id='2025-11-25'`. 

I noticed because I was writing an acceptance test that failed when I targeted #2025-08-15 but worked after I added `date-` at the start of the ID. 

Can't imagine how this would affect anyone but me, and either way, we certainly don't want to output invalid HTML. 

Thanks for considering.


jerclarke:
Ok ignore this for now, sorry I forgot to test what else would break and it does indeed break dragging of posts to a different date.

PR #754, opened 2025-08-26
Calendar caching issues: Refresh PR 548 from Cojennin for issue 505.

jerclarke:
Old out of date PR:
https://github.com/Automattic/Edit-Flow/pull/548

Issue this addresses:
https://github.com/Automattic/Edit-Flow/issues/505

This adapts the core elements of Cojennin's PR while integrating changes made to calendar.php since the patch was created.

The important thing it does is shift what we cache in `generate_post_li_html()` from being the final HTML to being the "information fields". This avoids having user-specific nonces get cached, allowing us to have the cache function without the user IDs, and thus allowing us to correctly purge the cache.

The issues with these changes is that they don't move the needle in terms of performance. On my local site, the old caching brought generation time from 3s to 1s. With this caching, there's no noticeable difference at all, on or off the results were between 3.09 and 3.11 seconds.


GaryJones:
Closing as superseded by #828, which removed the calendar `li_html_cache` entirely rather than attempting to fix the caching mechanism. Since the cache is now removed, this alternative caching approach is no longer needed.

PR #755, opened 2025-08-27
Disable calendar `li_html_cache` object cache entirely because it's uggy and the solution to performance problems isn't object caching

jerclarke:
Re: https://github.com/Automattic/Edit-Flow/issues/505

My analysis of this issue points to bugs in what is being loaded in the HTML, rather than database queries, as the cause of slowness in the calendar, so there's no reason to "save" the object cache which is currently totally broken (it is never invalidated, so out-of-date post info just hangs around until the cache clears itself eventually).

We should remove this cache ASAP so that the calendar is accurate, and deal with performance problems separately.


GaryJones:

Issue #756, opened 2025-08-27
Calendar: Make hierarchical taxonomies non-editable to solve performance and usability problems


jerclarke:
Following on from #505, this issue is to document the performance problem I discovered while investigating why the calendar `post_li_html` object cache was needed in the first place. 

[Note I make a lot of mistakes in this original ticket, see my reply below. The system does work, but badly and hierarchical taxonomies are the main problem that should be removed from 'editable' status.]

You can follow along with my research in [this reply](https://github.com/Automattic/Edit-Flow/issues/505#issuecomment-3226002443), but the upshot is that there's a whole useless system in the calendar based around the `editable` flag on the fields from `->get_post_information_fields()`, which doesn't seem to offer any functionality at this point, but has the effect of inserting enormous amounts of HTML processing time and page bloat. 

Here's how it works in the latest plugin version:

* `->get_post_information_fields()` is used to generate the list of fields that will be output in the Calendar's "inner information" for each post. i.e. the stuff that shows when you click the post to show details. 
* `->get_post_information_fields()` can insert a `['editable'] = true` flag on a field. Currently it **only** does this for taxonomies. 
* `->get_inner_information()`, which generates the post details HTML, checks this `editable` field and if it's `true`, loads a whole alternate version of the field, where it can be clicked to reveal a hidden form to edit the field. 
* `->get_inner_information()` calls `->get_editable_html()` to fetch the form specific to that field. 
* `->get_editable_html()` has various `$type`s inside it with custom form fields for each type, like text/location/date/user, but **ONLY** `taxonomy` and `taxonomy hierarchical` are actually used by the plugin itself. 
* `calendar.js` has a bunch of code to listen to these editable fields and send AJAX to update based on edits. 
* `calendar.css` has a few styles related to editable fields.

**The only built-in "editable field" in Edit Flow is taxonomies and they don't work at all.**

The thing about this is that the "editable taxonomy" fields just don't work. You end up with a `wp_dropdown_categories` pulldown menu to represent categories which both doesn't save anything, and would be disastrous if it did, because of course you are likely to have multiple categories already on the post. 

Here's what you see when you click the "editable" list of categories on a post in the Calendar:
<img width="510" height="347" alt="Image" src="https://github.com/user-attachments/assets/b7d06269-f923-490a-b419-14a12eade6e6" />

[...]

PR #757, opened 2025-08-27
Calendar: Remove 'taxonomy hierarchical' from editable fields for performance and usability

jerclarke:
Re: https://github.com/Automattic/Edit-Flow/issues/756

Stop hierarchical taxonomies like categories from being edited inside the calendar, while continuing to allow non-hierarchical taxonomies (tags) to be edited.

This avoids the horrific performance punishment on sites with many categories due to the full list of categories being inserted as a hidden field inside each and every post on the calendar.

It also avoids using a single-select field to update categories, which removes all categories but the one chosen when saved. Users are better off editing the post directly if they need to change the categories, at least unless and until we can offer a proper multi-select category picker in the calendar (a terrible idea because of the performance issues mentioned above).

### DIFF
--- a/modules/calendar/calendar.php
+++ b/modules/calendar/calendar.php
@@ -1102,12 +1102,6 @@ if ( ! class_exists( 'EF_Calendar' ) ) {
 				case 'taxonomy':
 					return '<input type="text" class="metadata-edit-' . esc_attr( $type ) . '" value="' . esc_attr( $value ) . '" />';
 				break;
-				case 'taxonomy hierarchical':
-					return wp_dropdown_categories( array(
-						'echo' => 0,
-						'hide_empty' => 0,
-					) );
-				break;
 			}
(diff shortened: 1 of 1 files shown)


GaryJones:

GaryJones:
While there may be users who only have single, non-hierachical taxonomies (category) in use, I agree that data loss is too risky. 

I think there is an opportunity to use Ajax to pull in hierarchical lists as needed, instead of loading up all of the options for all of the posts on the calendar screen at once. We'd want a bigger modal overlay experience to potentially contain multiple checkbox-based hierarchical  taxonomies.

jerclarke:
Thanks Gary! Sorry I didn't reply to the tickets last week, I was swamped with a work commitment and preparing a talk for a local WordPress meetup, of all things. 

I'll take a look at the others today!

PR #883, opened 2026-01-09
fix: prevent Edit Flow nonce checks from killing unrelated forms

### LINKED ISSUE #882: Cheatin' uh?
Whenever a post is saved, such as through a contact form submission, `transition_post_status` is fired off, and if that occurs outside of the wp-admin/edit.php screens (OR happens on those screens  but.. something or something... Unknown...) then edit-flow will fail a nonce check and kills the request.

https://github.com/Automattic/edit-flow/blob/1d0475490b755cf72120f8da27f0e626e0244ed0/modules/notifications/notifications.php#L603-L605

On WordCamp, this caused speaker contact form submissions to fail to save, because the nonce included was not for edit post, but rather for  speaker_submission or somesuch.

This doesn't seem to affect every use case.. oh, because if `_wpnonce` isn't set, this nonce check doesn't even properly run.

Ref: https://wordpress.slack.com/archives/C08M59V3P/p1767945920176379

Earlier issues seem to have been related to, and resolved by, 7ee2706ec680de9b67b420c99a6d65bd102d1e38 (Ref: https://wordpress.slack.com/archives/C08M59V3P/p1767902122649419 https://wordpress.slack.com/archives/C08M59V3P/p1767896599567679 )

Suggested solution:
 - Don't die like this in this handler; (edit: In any non-ajax handler)
 - Include your own nonce.
 - If nonce not set, OR nonce verification failed, return.
 - Verify this isn't systematic. `if ( ! empty( $_GET['_wpnonce'] ) && ! wp_verify_nonce` is rarely wanted. `if ( ! isset( $_POST['_wpnonce'] ) || ! wp_verify_nonce(` is usually what's wanted. Claude should know better.

GaryJones:
## Summary

Fixes #882

The notifications module was incorrectly checking the generic `_wpnonce` field against Edit Flow's expected action when posts were saved. This caused contact forms and other plugins that triggered post status transitions to fail, as Edit Flow would call `wp_die()` when their unrelated nonces didn't verify.

### Changes

1. **`save_post_subscriptions()`**: Now checks for Edit Flow's own form indicator (`ef-save_followers`) first, then verifies its own dedicated nonce field (`ef_notifications_nonce`) with action `save_user_usergroups`. Returns early instead of calling `wp_die()`.

2. **`handle_user_post_subscription()`**: Fixed faulty nonce logic that allowed requests without any nonce to pass through. Changed from `! empty() && ! verify()` to `! isset() || ! verify()`.

---

### The Scenario

A user submits a **speaker contact form** on WordCamp. This form:
1. Has its own `$_POST['_wpnonce']` field for action `speaker_submission`
2. Creates/updates a post behind the scenes (to store the speaker data)
3. This triggers WordPress's `transition_post_status` hook

### Before (The Bug)

```php
public function save_post_subscriptions( $new_status, $old_status, $post ) {
    // Skip revisions...
    
    if ( ! empty( $_POST['_wpnonce'] ) && ! wp_verify_nonce( $_POST['_wpnonce'], 'update-post_' . $post->ID ) ) {
        $this->print_ajax_response( 'error', ... );  // This calls wp_die()!
    }
    
    if ( isset( $_POST['ef-save_followers'] ) && ... ) {
        // Save followers
    }
}
```

**What happened:**
1. Contact form submits with `$_POST['_wpnonce']` = nonce for `speaker_submission`
2. Post is created → `transition_post_status` fires → `save_post_subscriptions()` runs
3. `! empty( $_POST['_wpnonce'] )` → **TRUE** (the contact form set it)
4. `! wp_verify_nonce( $_POST['_wpnonce'], 'update-post_123' )` → **TRUE** (wrong action, verification fails)
5. Both conditions true → `print_ajax_response()` → `wp_die()` → **Request killed**
6. Speaker contact form submission fails with "Cheatin' uh?"

### After (The Fix)

```php
public function save_post_subscriptions( $new_status, $old_status, $post ) {
    // Skip revisions...
    
    // Only process if Edit Flow's form was submitted
    if ( ! isset( $_POST['ef-save_followers'] ) ) {
        return;  // ← Exit early, don't interfere
    }
    
    // Check capability
    if ( ! current_user_can( $this->edit_post_subscriptions_cap ) ) {
        return;
    }
    
    // Verify Edit Flow's OWN nonce
    if ( ! isset( $_POST['ef_notifications_nonce'] ) || ! wp_verify_nonce( $_POST['ef_notifications_nonce'], 'save_user_usergroups' ) ) {
        return;  // ← Return, don't die
    }
    
    // Save followers...
}
```

**What happens now:**
1. Contact form submits with `$_POST['_wpnonce']` = nonce for `speaker_submission`
2. Post is created → `transition_post_status` fires → `save_post_subscriptions()` runs
3. `! isset( $_POST['ef-save_followers'] )` → **TRUE** (contact form didn't set this)
4. Function returns early → **No interference**
5. Contact form submission succeeds ✓

### Key changes

- **Check for Edit Flow's form first** (`ef-save_followers`) before doing anything
- **Use Edit Flow's own nonce field** (`ef_notifications_nonce`) instead of the generic `_wpnonce`
- **Return instead of die** - even if something fails, don't kill the entire request

---

## Test plan

- [x] All existing notifications tests pass
- [x] New tests added for:
  - Accepting Edit Flow's own nonce
  - Ignoring requests without Edit Flow form data
  - Not dying when unrelated form nonce is present (the #882 scenario)
  - Returning early with invalid nonce (not dying)
  - AJAX handler properly requires nonce

🤖 Generated with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>

### DIFF
--- a/modules/notifications/notifications.php
+++ b/modules/notifications/notifications.php
@@ -554,8 +554,9 @@ if ( ! class_exists( 'EF_Notifications' ) ) {
 		 * @since 0.8
 		 */
 		public function handle_user_post_subscription() {
+			// Require a valid nonce for this AJAX request.
 			// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized -- Nonce value passed directly to wp_verify_nonce().
-			if ( ! empty( $_GET['_wpnonce'] ) && ! wp_verify_nonce( $_GET['_wpnonce'], 'ef_notifications_user_post_subscription' ) ) {
+			if ( ! isset( $_GET['_wpnonce'] ) || ! wp_verify_nonce( $_GET['_wpnonce'], 'ef_notifications_user_post_subscription' ) ) {
 				$this->print_ajax_response( 'error', $this->module->messages['nonce-failed'] );
 			}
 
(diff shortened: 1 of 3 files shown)


dd32:
@GaryJones A quick visual review seems good to me.

However, I'm questioning if this is correct:
> `			add_action( 'transition_post_status', [ $this, 'save_post_subscriptions' ], 0, 3 );`

I think it might be better if that was hooked onto the [save_post](https://developer.wordpress.org/reference/hooks/save_post/) action instead of the transition.

GaryJones:
Thanks for the review @dd32!

The `transition_post_status` hook is intentional here — `save_post_subscriptions` runs at priority 0 before `notification_status_change` runs at priority 10 on the same hook. Since `save_post` fires *after* `transition_post_status` in WordPress core, switching just this function would break that ordering.

That said, your instinct is sound — `save_post` is the more canonical hook for saving meta alongside a post. I've opened #886 to track the potential refactor (moving both functions to `save_post`), as it would be a larger change with potential breaking implications that's outside the scope of this bug fix.

PR #980, opened 2026-06-04
update EF_Module->users_select_form() to remove faulty escaping of checkbox attributes and fix "No one will be notified" message

jerclarke:
Removes `esc_attr()` from use of `$checked` and `$current_user_class` variables before display.

Using `esc_html()` on these values destroys the quotations inside them, resulting in HTML output like this in the HTML source:

```html
checked=&quot;checked&quot;class=&quot;post_following_list-current_user&quot;
```

Notably, this leaves `checked` functional, because the browser interprets all that mess as `true`, so the basics of subscribing and unsubscribing people continues to function superficially, but of course the `class` is non-functional.

The bug that made me notice this was that the `$current_user_class` wasn't having the intended effect in Editorial Comments, where it should ensure that the current user doesn't show up as a recipient of editorial comments. Without this patch, you always see yourself listed as a recipient, and you will NEVER see the "No one will be notified" message.

I don't know who added that escaping but it seems unnecessary to me, considering the HTML being output is defined deterministically in the code right above the usage, so there's no way it could be suspicious. Honestly the way those lines were formatted makes me think it was a misbehaving AI. I reformatted them to match how the rest of the `<input>` is output.

I'm not familiar with the testing framework for EF, but it would be great to have some kind of test to confirm that "No one will be notified" shows up at the appropriate time in the UI, to avoid this kind of thing getting out in the future.

Thanks for reviewing!

## Steps to Test

* Open up a post, save, and reload so that Editorial Comments are functioning
* Click to start a new response
* Without the patch: 
  * See yourself listed as a recipient
  * HTML source shows mess of `&quot`
* With the patch: 
  * See "No one will be notified" until you add a user who is not yourself
  * HTML source shows clean `checked` and `class` attributes.

### DIFF
--- a/common/php/class-module.php
+++ b/common/php/class-module.php
@@ -560,15 +560,22 @@ if ( ! class_exists( 'EF_Module' ) ) {
 			<ul class="<?php echo esc_attr( $list_class ); ?>">
 				<?php
 				foreach ( $users as $user ) :
-					$checked = ( in_array( $user->ID, $selected ) ) ? 'checked="checked"' : '';
-					// Add a class to checkbox of current user so we know not to add them in notified list during notifiedMessage() js function.
-					$current_user_class = ( get_current_user_id() == $user->ID ) ? 'class="post_following_list-current_user" ' : '';
+					$checked = in_array( $user->ID, $selected );
 					?>
 					<li>
 						<label for="<?php echo esc_attr( $input_id . '-' . $user->ID ); ?>">
 							<div class="ef-user-subscribe-actions">
 								<?php do_action( 'ef_user_subscribe_actions', $user->ID, $checked ); ?>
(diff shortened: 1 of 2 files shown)


jerclarke:
Ok, I see that the CI is complaining about this even though there's no real problem. The "solution" in the current version breaks the actual behavior so I guess another fix is needed?

I had already tried using `esc_html()` but it had the same effect as `esc_attr()`. 

Can you update this to satisfy the checker? This code is spaghetti so there's a million ways to clean it up. 

I'm trying to get codesniffer set up in my copy of EF but have to step away before finding a fix that satisfies the linter, so I figure I'd let you know I understand the problem at least.

jerclarke:
Ok, updated PR changes things more than I'd want but addresses the PHPCS complaints without breaking anything, as far as I can tell 🫡

jerclarke:
Thanks Gary! Appreciate the extra test too 🙏🏻

PR #1016, opened 2026-06-22
fix: escape calendar item action link hrefs with esc_url() and link text with esc_html__()

thisismyurl:
\`EF_Calendar::get_inner_information()\` builds five action links for each calendar item — Edit, Trash, Preview, View, and Save. Three of those hrefs came from functions that return plain URLs (\`get_edit_post_link()\`, \`get_delete_post_link()\`, \`get_permalink()\`), and all five link text strings used \`__()\` directly. Neither set went through an escape function.

PR #993 made the same fix for \`story-budget.php\`; the calendar module's \`item_actions\` block was the parallel section that was missed.

This PR wraps the three hrefs in \`esc_url()\` and changes all five link-text calls to \`esc_html__()\`. The Preview branch already had \`esc_url()\` on its href — only the link text needed updating there.

The new test class \`CalendarEscapingTest\` injects a URL with a bare \`&\` via a WordPress filter (\`get_edit_post_link\`, \`post_link\`) and asserts the output contains \`&#038;\` (what \`esc_url()\` produces for \`&\`) rather than the raw \`&\`. A separate assertion confirms no raw \`&\` survives in any href attribute. \`test_calendar_view_link_href_is_url_escaped\` uses a published post so \`get_permalink()\` is the active code path (not the preview branch). The trash link is omitted from integration coverage because \`get_delete_post_link()\` routes through \`wp_nonce_url()\` which pre-encodes \`&\` as \`&amp;\` — an integration test for that path cannot distinguish "esc_url() ran" from "wp_nonce_url() pre-encoded the value."

Tests reviewed by inspection; not executed locally (no integration harness in my environment) — relying on CI for the first run.

No visual or interaction change; escaping only, link text and structure unchanged, no a11y impact.

(props @thisismyurl)

(full disclosure: AI helped me identify the issue and verify my work)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ### Security
 
-* fix: escape calendar item action link hrefs with esc_url() and link text with esc_html__() by @thisismyurl
+* Escape calendar item action link hrefs with esc_url() and link text with esc_html__() (props @thisismyurl)
 
 ## [0.11.0] - 2026-06-10
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased] - TBD
+
+### Security
+
+* fix: escape calendar item action link hrefs with esc_url() and link text with esc_html__() by @thisismyurl
+
 ## [0.11.0] - 2026-06-10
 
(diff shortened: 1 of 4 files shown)


GaryJones:
Reviewed — fix is correct, CI green. The escaping is right and the mutation-style tests (assert `&#038;`, deny raw `&`) are a solid regression guard; good call documenting why the trash path can't be covered at the integration layer.

One scope question: the `item_actions` block is fully handled, but two other spots still pass a URL into an `href` without `esc_url()` — `calendar.php:110` and `story-budget.php:105`, both `admin_url()` inside a translated `short_description`. They're static/safe so not blocking, but if this is meant to close out the escaping sweep rather than just this block, worth folding in (here or a follow-up).

thisismyurl:
Thanks @GaryJones — folded both in (63b1e7d). `calendar.php` and `story-budget.php` now wrap their `admin_url()` short-description hrefs with `esc_url()`, and the changelog has a matching entry. That should close out the escaping sweep for this surface rather than leaving a follow-up dangling.

